PR #16, opened 2021-06-24
SigV4_GenerateHTTPAuthorization() API Functionality

sukhmanm:
*Description of changes:* Adds API functionality for SigV4_GenerateAuthorizationHeader(). The unit tests for this function will be added in a separate PR, hence the failing check here.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


source/sigv4.c
```diff
@@ -437,6 +562,398 @@ static SigV4Status_t parseDate( const char * pDate,
...
+
+        canonicalRequest->pBufCur += remainingLen;
+        *( canonicalRequest->pBufCur++ ) = '\n';
+
+        canonicalRequest->bufRemaining -= remainingLen + 1;
+    }
+
+/*-----------------------------------------------------------*/
+
+    static void generateCanonicalQuery( const char * pQuery,
```
gshvang, line 796: It will be better if more comments can be added to the function to understand it easily.

gshvang (commented):
In General, more comments can be added to the code in the functions to understand easily how they are achieving the required functionality.

source/sigv4.c
```diff
@@ -436,6 +561,398 @@ static SigV4Status_t parseDate( const char * pDate,
 
 /*-----------------------------------------------------------*/
 
+/* Converts a hex character to its integer value */
+static char hexToInt( char pHex )
```
gshvang, line 565: Declaration of this static function is missing.

source/sigv4.c
```diff
@@ -436,6 +561,398 @@ static SigV4Status_t parseDate( const char * pDate,
...
+        byte = byte - 'A' + 10;
+    }
+
+    return byte & 0xF;
+}
+
+/*-----------------------------------------------------------*/
+
+/* Hex digest of provided parameter string. */
+static void hexEncode( SigV4String_t * pInputStr,
```
gshvang, line 588: Declaration of this static function is missing.

source/sigv4.c
```diff
@@ -436,6 +561,398 @@ static SigV4Status_t parseDate( const char * pDate,
...
+                       SigV4String_t * pHexOutput )
+{
+    static const unsigned char digitArr[] = "0123456789abcdef";
```
gshvang, line 591: Its better to assert the input parameters here in this function.

source/sigv4.c
```diff
@@ -436,6 +561,398 @@ static SigV4Status_t parseDate( const char * pDate,
...
+        *( hex++ ) = digitArr[ ( pInputStr->pData[ i ] & 0xF0 ) >> 4 ];
+        *( hex++ ) = digitArr[ ( pInputStr->pData[ i ] & 0xF0 ) ];
+    }
+
+    pHexOutput->dataLen = pInputStr->dataLen * 2;
+}
+
+/*-----------------------------------------------------------*/
+
+static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
```
gshvang, line 606: Assertion of input parameters should be done for this function.

gshvang (changes requested):

source/sigv4.c
```diff
@@ -436,6 +562,407 @@ static SigV4Status_t parseDate( const char * pDate,
...
+}
+
+/*-----------------------------------------------------------*/
+
+/* Hex digest of provided parameter string. */
+static void hexEncode( SigV4String_t * pInputStr,
+                       SigV4String_t * pHexOutput )
+{
+    static const unsigned char digitArr[] = "0123456789abcdef";
+    unsigned char * hex = pHexOutput->pData;
```
gshvang, line 593: Assertion should be done before accessing pHexOutput. This should also be asserted pHexOutput->pData

source/sigv4.c
```diff
@@ -436,6 +562,407 @@ static SigV4Status_t parseDate( const char * pDate,
...
+    pHexOutput->dataLen = pInputStr->dataLen * 2;
+}
+
+/*-----------------------------------------------------------*/
+
+static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
+                                         SigV4String_t * pCredScope )
+{
+    SigV4Status_t returnVal = SigV4InvalidParameter;
+    unsigned char * pBufWrite = pCredScope->pData;
```
gshvang, line 615: Assertion should be done before accessing pCredScope.

gshvang (changes requested):

source/sigv4.c
```diff
@@ -436,6 +561,392 @@ static SigV4Status_t parseDate( const char * pDate,
...
+
+        return ( pInput->pData == NULL || pInput->dataLen == 0 ) ? returnVal : !returnVal;
+    }
+
+/*-----------------------------------------------------------*/
+
+    static int cmpKeyValue( const void * pFirstVal,
+                            const void * pSecondVal )
+    {
+        SigV4KeyValuePair_t * pFirst = ( SigV4KeyValuePair_t * ) pFirstVal;
```
gshvang, line 683: Same here assertion should be before assignment.

source/sigv4.c
```diff
@@ -436,6 +561,392 @@ static SigV4Status_t parseDate( const char * pDate,
...
+/*-----------------------------------------------------------*/
+
+    static void encodeURI( const char * pURI,
+                           size_t uriLen,
+                           char * pCanonicalURI,
+                           size_t * canonicalURILen,
+                           bool encodeSlash,
+                           bool nullTerminate )
+    {
+        const char * pURILoc = pURI;
```
gshvang, line 715: Assertion should be done before assignment.

source/sigv4.c
```diff
@@ -436,6 +561,392 @@ static SigV4Status_t parseDate( const char * pDate,
...
+    }
+
+/*-----------------------------------------------------------*/
+
+    static void generateCanonicalURI( const char * pURI,
+                                      size_t uriLen,
+                                      bool encodeOnce,
+                                      canonicalContext_t * canonicalRequest )
+    {
+        size_t encodedLen, remainingLen = canonicalRequest->bufRemaining;
```
gshvang, line 764: Assertion should be done before assigning the values.

source/sigv4.c
```diff
@@ -436,6 +561,392 @@ static SigV4Status_t parseDate( const char * pDate,
...
+/*-----------------------------------------------------------*/
+
+    static void generateCanonicalQuery( const char * pQuery,
+                                        size_t queryLen,
+                                        canonicalContext_t * canonicalRequest )
+    {
+        size_t index = 0U;
+        size_t i = 0U;
+        size_t remainingLen = canonicalRequest->bufRemaining;
+        char * pBufLoc = canonicalRequest->pBufCur;
```
gshvang, line 795: Assertion for canonicalRequest->pBufCur is missing

source/sigv4.c
```diff
@@ -436,6 +561,401 @@ static SigV4Status_t parseDate( const char * pDate,
...
+    }
+
+/*-----------------------------------------------------------*/
+
+    static void generateCanonicalQuery( const char * pQuery,
+                                        size_t queryLen,
+                                        canonicalContext_t * canonicalRequest )
+    {
+        size_t index, i = 0U;
+        size_t remainingLen = canonicalRequest->bufRemaining;
```
gshvang, line 801: Assertion for canonicalRequest->bufRemaining is missing.

source/sigv4.c
```diff
@@ -436,6 +561,401 @@ static SigV4Status_t parseDate( const char * pDate,
...
+
+/*-----------------------------------------------------------*/
+
+    static void generateCanonicalURI( const char * pURI,
+                                      size_t uriLen,
+                                      bool encodeOnce,
+                                      canonicalContext_t * canonicalRequest )
+    {
+        char * pBufLoc = NULL;
+        size_t encodedLen, remainingLen = canonicalRequest->bufRemaining;
```
gshvang, line 771: Assertion for canonicalRequest->bufRemaining is missing.

gshvang (dismissed):

source/sigv4.c
```diff
@@ -436,6 +561,402 @@ static SigV4Status_t parseDate( const char * pDate,
...
+        }
+
+        canonicalRequest->pBufCur = pBufLoc;
+    }
+
+#endif /* #if ( SIGV4_USE_CANONICAL_SUPPORT == 1 ) */
+
+/*-----------------------------------------------------------*/
+
+static SigV4Status_t verifySigV4Parameters( const SigV4Parameters_t * pParams )
```
cobusve, line 865: I have 2 comments here, firstly a lot of error strings like this can end up taking a large amount of memory on smaller systems. Did we consider if it is possible to implement this in such a way that we can minimize the memory footprint on smaller devices?

Second question, it is always useful when you return an error code which represents a the same granularity of knowledge about the problem that you know, that way it is possible for callers to act accordingly when they get an error . In this case we return the same error code for any parameter which is incorrect.  If we decide to just return SigV4InvalidParameter then I would get rid of all the duplicated assignments and invert the logic to set the return status to SigV4InvalidParameter at init and just set it to success in a last else?

cobusve: One further question, this is a static function defined in this file but never called?

source/sigv4.c
```diff
@@ -436,6 +561,402 @@ static SigV4Status_t parseDate( const char * pDate,
...
+        assert( *canonicalURILen > 0U );
+
+        pURILoc = ( const char * ) pURI;
+        pBufLoc = ( char * ) pCanonicalURI;
+
+        while( index < uriLen && *pURILoc )
+        {
+            if( isalnum( *pURILoc ) || ( *pURILoc == '-' ) || ( *pURILoc == '_' ) || ( *pURILoc == '.' ) || ( *pURILoc == '~' ) )
+            {
+                *pBufLoc++ = *pURILoc;
```
cobusve, line 734: Please use brackets on this post increment to make it clear if we are intending to increment the pointer or the value it is pointing to.

source/sigv4.c
```diff
@@ -436,6 +561,402 @@ static SigV4Status_t parseDate( const char * pDate,
...
+                index++;
+            }
+            else if( ( *pURILoc == '/' ) && !encodeSlash )
+            {
+                *pBufLoc++ = *pURILoc;
```
cobusve, line 739: Please use brackets on this post increment to make it clear if we are intending to increment the pointer or the value it is pointing to.

source/sigv4.c
```diff
@@ -436,6 +561,402 @@ static SigV4Status_t parseDate( const char * pDate,
...
+                index++;
+            }
+            else
+            {
+                *pBufLoc++ = '%';
```
cobusve, line 744: Please use brackets on this post increment to make it clear if we are intending to increment the pointer or the value it is pointing to.

source/sigv4.c
```diff
@@ -436,6 +561,402 @@ static SigV4Status_t parseDate( const char * pDate,
...
+
+                index += 3;
+            }
+
+            pURILoc++;
+        }
+
+        if( nullTerminate )
+        {
+            *pBufLoc++ = '\0';
```
cobusve, line 756: Please use brackets on this post increment to make it clear if we are intending to increment the pointer or the value it is pointing to.

and so on I am going to leave it up to you to find the rest

sukhmanm: Thanks, updating these now, alongside a few minor changes in relation to your earlier comment, and will follow up for review on those shortly.

source/sigv4.c
sukhmanm: That is an interesting point, and something we hadn't given too much thought. Although the logging approach used was modelled a bit after coreHTTP ([`HTTPClient_ReadHeader`](https://github.com/FreeRTOS/coreHTTP/blob/4c896b98d7d9bfe2868e00982f2eaf64ec44752f/source/core_http_client.c#L2496-L2558), for example), it does seem like there could be much room for optimization here, just given the number of null check repetitions required, as one example. 

As you alluded to, the general approach we followed here was to supply detailed error logs explaining the rather vague status values returned from the library, but it might be worth expanding the range and detail of these enums, and maybe providing some sort of generic function to log them and avoid redundancy as much as possible. 

I will discuss this in some more detail and if time permits, address any refactoring decided upon independently. Let me know if you disagree with this approach or have any further suggestions on this. Thanks.

source/sigv4.c
```diff
@@ -436,6 +561,402 @@ static SigV4Status_t parseDate( const char * pDate,
...
+    assert( pHexOutput != NULL );
+    assert( pInputStr->pData != NULL );
+    assert( pHexOutput->pData != NULL );
+
+    hex = pHexOutput->pData;
+
+    for( i = 0; i < pInputStr->dataLen; i++ )
+    {
+        *( hex++ ) = digitArr[ ( pInputStr->pData[ i ] & 0xF0 ) >> 4 ];
+        *( hex++ ) = digitArr[ ( pInputStr->pData[ i ] & 0xF0 ) ];
```
emilmelnikov, line 582:
```suggestion
        *( hex++ ) = digitArr[ ( pInputStr->pData[ i ] & 0x0F ) ];
```

source/sigv4.c
```diff
@@ -436,6 +561,402 @@ static SigV4Status_t parseDate( const char * pDate,
...
+    /* Use only the first 8 characters from the provided ISO 8601 string (YYYYMMDD). */
+    bytesWritten = snprintf( ( char * ) pBufWrite,
+                             ISO_DATE_SCOPE_LEN + 1,
+                             "%*s",
+                             ISO_DATE_SCOPE_LEN,
+                             pSigV4Params->pDateIso8601 );
+
+    if( bytesWritten == ISO_DATE_SCOPE_LEN )
+    {
+        bytesWritten = snprintf( ( char * ) pBufWrite,
```
aggarg, line 612: `pBufWrite` is set to `pCredScope->pData;` - won't multiple `snprintf` calls overwrite the data or am I missing something?

source/sigv4.c
```diff
@@ -436,6 +561,402 @@ static SigV4Status_t parseDate( const char * pDate,
...
+
+    assert( pSigV4Params != NULL );
+    assert( pCredScope != NULL );
+    assert( pCredScope->pData != NULL );
+
+    pBufWrite = pCredScope->pData;
+
+    /* Use only the first 8 characters from the provided ISO 8601 string (YYYYMMDD). */
+    bytesWritten = snprintf( ( char * ) pBufWrite,
+                             ISO_DATE_SCOPE_LEN + 1,
```
aggarg, line 605: Shouldn't this be `pCredScope->dataLen` to avoid buffer overflow?

source/sigv4.c
```diff
@@ -436,6 +561,402 @@ static SigV4Status_t parseDate( const char * pDate,
...
+        bool returnVal = true;
+
+        assert( pInput != NULL );
+
+        return ( pInput->pData == NULL || pInput->dataLen == 0 ) ? returnVal : !returnVal;
+    }
+
+/*-----------------------------------------------------------*/
+
+    static int cmpKeyValue( const void * pFirstVal,
```
aggarg, line 680: why use `void*` here? Can we not use `SigV4KeyValuePair_t`?

source/sigv4.c
```diff
@@ -436,6 +561,402 @@ static SigV4Status_t parseDate( const char * pDate,
...
+            else if( ( *pURILoc == '/' ) && !encodeSlash )
+            {
+                *( pBufLoc++ ) = *pURILoc;
+                index++;
+            }
+            else
+            {
+                *( pBufLoc++ ) = '%';
+                *( pBufLoc++ ) = *pURILoc >> 4;
+                *( pBufLoc++ ) = *pURILoc & 15;
```
aggarg, line 746:
```suggestion
                *( pBufLoc++ ) = *pURILoc & 0x0F;
```

source/sigv4.c
```diff
@@ -436,6 +561,402 @@ static SigV4Status_t parseDate( const char * pDate,
...
+    {
+        char * pBufLoc = NULL;
+        size_t encodedLen, remainingLen = 0U;
+
+        assert( pURI != NULL );
+        assert( canonicalRequest != NULL );
+        assert( canonicalRequest->pBufCur != NULL );
+
+        pBufLoc = ( char * ) canonicalRequest->pBufCur;
+        encodedLen, remainingLen = canonicalRequest->bufRemaining;
```
aggarg, line 778: Is `remainingLen` needed in this line?

source/sigv4.c
sukhmanm: Thanks

source/sigv4.c
sukhmanm: It was just to match the function signature that's defined for the `qsort()` comparison functions, which is the intention for this function as well (during canonicalization).

aggarg (approved):
The remaining comments will be addressed in a separate PR.

abhidixi11:
See comment from @aggarg

PR #28, opened 2021-08-02
Add Canonical and signed headers functions

gshvang:
This PR adds functions to generate canonical and signed headers needed to generate the signature.


source/sigv4.c
```diff
@@ -793,6 +841,231 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                }
+                else
+                {
+                    end++;
+                }
+            }
+        }
+
+        /* Sorting headers based on keys. */
+        qsort( canonicalRequest->pHeadersLoc, 5, sizeof( SigV4KeyValuePair_t ), cmpKeyValue );
```
yourslab, line 1055: The number of keys here are fixed

yourslab: The function name `cmpKeyValue` is a little misleading as the headers only need to be sorted by field
```suggestion
        qsort( canonicalRequest->pHeadersLoc, 5, sizeof( SigV4KeyValuePair_t ), cmpField );
```

source/sigv4.c
```diff
@@ -793,6 +841,231 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                    fieldFlag = 0;
+                }
+                else
+                {
+                    end++;
+                }
+            }
+            else
+            {
+                if( pHeaders[ i ] == '\n' )
```
yourslab, line 1038: `\r\n` is the start of a new header. The value of an HTTP header is allowed to have `\n`, so that's a corner case to watch out for.

source/sigv4.c
```diff
@@ -793,6 +841,231 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        char * pBufLoc;
+        SigV4Status_t sigV4Status = SigV4Success;
+
+        assert( canonicalRequest != NULL );
+        assert( canonicalRequest->pBufCur != NULL );
+
+        pBufLoc = canonicalRequest->pBufCur;
+        buffRemaining = canonicalRequest->bufRemaining;
+
+        for( noOfHeaders = 0; noOfHeaders < SIGV4_MAX_HTTP_HEADER_COUNT; noOfHeaders++ )
```
yourslab, line 891: I think it's better to have an actual count of the headers that you can pass to this function to avoid unnecessary iterations.

source/sigv4.c
```diff
@@ -793,6 +841,231 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        {
+            if( canonicalRequest->pHeadersLoc[ noOfHeaders ].key.pData != NULL )
```
yourslab, line 893: This should never really happen, right? This is more appropriately handled by an `assert`. Note that an asserts are stripped away after compiling in release mode. The purpose of an assert is just to help with debugging.

source/sigv4.c
```diff
@@ -793,6 +841,231 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].value.pData = start;
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].value.dataLen = ( end - start );
+                    end++;
+                    start = end;
+                    fieldFlag = 1;
+                    noOfHeaders++;
+                }
+                else
+                {
+                    end++;
```
yourslab, line 1049: Nit: `end++` is written in every branch. It's possible to just do it once and use `end + 1U` where needed.

source/sigv4.c
```diff
@@ -793,6 +841,231 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+            *pBufLoc = ':';
+            pBufLoc++;
+            value = canonicalRequest->pHeadersLoc[ headerNum ].value.pData;
+            trimValueLen = 0;
+
+            for( i = 0; i < valLen; i++ )
+            {
+                if( isspace( value[ i ] ) && ( i + 1 <= valLen ) && ( ( i + 1 == valLen ) || isspace( value[ i + 1 ] ) || ( trimValueLen == 0 ) ) )
+                {
+                    continue;
```
yourslab, line 950: MISRA forbids use of `continue` :(

It's a guideline though, and I don't always agree with it. You can just use De Morgan's law here though to avoid use of `continue`.

source/sigv4.c
```diff
@@ -793,6 +841,231 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                    *pBufLoc = canonicalRequest->pHeadersLoc[ headerNum ].value.pData[ i ];
+                    pBufLoc++;
+                    trimValueLen++;
+                }
+            }
+
+            *pBufLoc = '\n';
+            pBufLoc++;
+
+            buffRemaining -= ( keyLen + trimValueLen + 2 );
```
yourslab, line 963: This looks like it may be susceptible to overflow. CBMC should catch it though.

source/sigv4.c
```diff
@@ -793,6 +841,231 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                }
+            }
+        }
+
+        return sigV4Status;
+    }
+
+    static SigV4Status_t generateCanonicalHeaders( const char * pHeaders,
+                                                   size_t headersLen,
+                                                   CanonicalContext_t * canonicalRequest )
```
yourslab, line 1002: If `SIGV4_HTTP_HEADERS_ARE_CANONICAL_FLAG` is activated in `pParams->flags`, then we must scan the canonicalized headers provided by the user in order to write the signed headers. The format is actually very similar, but the difference is that you no longer need to sort the headers if the canonicalized headers are provided.

source/sigv4.c
gshvang: It will not overflow as before writing to buffer I have checked that there should be enough buffer to accommodate the data.`if( keyLen + valLen + 2 > buffRemaining )
        {
            return SigV4InsufficientMemory;
        }`

source/sigv4.c
```diff
@@ -793,6 +854,232 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        assert( canonicalRequest->pBufCur != NULL );
+
+        pBufLoc = canonicalRequest->pBufCur;
+        buffRemaining = canonicalRequest->bufRemaining;
+
+        keyLen = canonicalRequest->pHeadersLoc[ headerNum ].key.dataLen;
+
+        if( keyLen + 1 > buffRemaining )
+        {
+            return SigV4InsufficientMemory;
```
yourslab, line 875: MISRA only allows a single `return` statement

source/sigv4.c
```diff
@@ -793,6 +854,232 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+            {
+                if( pHeaders[ i ] == ':' )
+                {
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].key.pData = start;
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].key.dataLen = ( end - start );
+                    start = end + 1;
+                    fieldFlag = 0;
+                }
+
+                end + 1U;
```
yourslab, line 1048: I don't think this actually increments it. You can also put this outside the if statement and at the end of the for loop instead.

source/sigv4.c
```diff
@@ -793,6 +854,232 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+            }
+
+            /* Extracting each header key and value from the headers string. */
+            if( fieldFlag == 1 )
+            {
+                if( pHeaders[ i ] == ':' )
+                {
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].key.pData = start;
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].key.dataLen = ( end - start );
+                    start = end + 1;
```
yourslab, line 1044:
```suggestion
                    start = end + 1U;
```

source/sigv4.c
```diff
@@ -677,8 +738,8 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
 
 /*-----------------------------------------------------------*/
 
-    static int cmpKeyValue( const void * pFirstVal,
-                            const void * pSecondVal )
+    static int cmpField( const void * pFirstVal,
```
yourslab, line 741: MISRA requires us to always specify the bit-width of integral types like `int`
```suggestion
    static int32_t cmpField( const void * pFirstVal,
```

source/sigv4.c
```diff
@@ -793,6 +854,232 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].key.dataLen = ( end - start );
+                    start = end + 1;
+                    fieldFlag = 0;
+                }
+
+                end + 1U;
+            }
+            else
+            {
+                if( ( pHeaders[ i ] == '\r' ) && ( ( i + 1 ) < headersLen ) && ( pHeaders[ i + 1 ] == '\n' ) )
```
yourslab, line 1052: For precanonicalized headers provided by customer, they are only separated by `\n` instead of `\r\n`. We should 
handle that case as well.

yourslab (approved):

source/sigv4.c
```diff
@@ -793,6 +854,230 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                {
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].key.pData = start;
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].key.dataLen = ( end - start );
+                    start = end + 1U;
+                    fieldFlag = 0;
+                }
+            }
+            else
+            {
+                if( ( ( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) && ( pHeaders[ i ] == '\n' ) ) || ( !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) && ( pHeaders[ i ] == '\r' ) && ( ( i + 1 ) < headersLen ) && ( pHeaders[ i + 1 ] == '\n' ) ) )
```
yourslab, line 1050: You only need if..elseif statement to simplify the complexity here.

source/sigv4.c
```diff
@@ -793,6 +854,230 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+    {
+        size_t noOfHeaders = 0, keyLen = 0, i = 0;
+        SigV4Status_t sigV4Status = SigV4Success;
+
+        assert( canonicalRequest != NULL );
+        assert( canonicalRequest->pBufCur != NULL );
+
+        for( noOfHeaders = 0; noOfHeaders < headerCount; noOfHeaders++ )
+        {
+            assert( ( canonicalRequest->pHeadersLoc[ noOfHeaders ].key.pData ) != NULL )
```
yourslab, line 915:
```suggestion
            assert( ( canonicalRequest->pHeadersLoc[ noOfHeaders ].key.pData ) != NULL );
```

source/sigv4.c
```diff
@@ -793,6 +854,230 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+    {
+        size_t noOfHeaders = 0;
+        SigV4Status_t sigV4Status = SigV4Success;
+
+        assert( canonicalRequest != NULL );
+        assert( canonicalRequest->pBufCur != NULL );
+
+        for( noOfHeaders = 0; noOfHeaders < headerCount; noOfHeaders++ )
+        {
+            if( canonicalRequest->pHeadersLoc[ noOfHeaders ].key.pData != NULL )
```
yourslab, line 1000: This should also be asserted

source/sigv4.c
```diff
@@ -793,6 +854,230 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+            }
+
+            *pBufLoc = ':';
+            pBufLoc++;
+            value = canonicalRequest->pHeadersLoc[ headerNum ].value.pData;
+            trimValueLen = 0;
+
+            for( i = 0; i < valLen; i++ )
+            {
+                if( !( isspace( value[ i ] ) && ( i + 1 <= valLen ) && ( ( i + 1 == valLen ) || isspace( value[ i + 1 ] ) || ( trimValueLen == 0 ) ) ) )
```
yourslab, line 969: > To create the canonical headers list, convert all header names to lowercase and remove leading spaces and trailing spaces. Convert sequential spaces in the header value to a single space.

I think this piece of code just removes all spaces from the string and needs more testing

source/sigv4.c
```diff
@@ -793,6 +854,230 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        if( ( sigV4Status == SigV4Success ) && !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) )
+        {
+            qsort( canonicalRequest->pHeadersLoc, noOfHeaders, sizeof( SigV4KeyValuePair_t ), cmpField );
+        }
+
+        if( ( sigV4Status == SigV4Success ) && !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) )
+        {
+            sigV4Status = appendCanonicalizedHeaders( noOfHeaders, canonicalRequest );
+        }
+
```
yourslab, line 1073: There must be a newline between the canonicalized headers and signed headers.

yourslab (commented):
Found some issues while integrating with my branch

source/sigv4.c
```diff
@@ -793,6 +854,230 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                    fieldFlag = 0;
+                }
+            }
+            else
+            {
+                if( ( ( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) && ( pHeaders[ i ] == '\n' ) ) || ( !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) && ( pHeaders[ i ] == '\r' ) && ( ( i + 1 ) < headersLen ) && ( pHeaders[ i + 1 ] == '\n' ) ) )
+                {
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].value.pData = start;
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].value.dataLen = ( end - start );
+                    start = end + 1U;
```
yourslab, line 1054: Note that this includes the `\n` at the start of the key.

source/sigv4.c
```diff
@@ -793,6 +854,230 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        {
+            sigV4Status = SigV4InsufficientMemory;
+        }
+        else
+        {
+            for( i = 0; i < keyLen; i++ )
+            {
+                if( !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) )
+                {
+                    *pBufLoc = tolower( canonicalRequest->pHeadersLoc[ headerNum ].key.pData[ i ] );
```
yourslab, line 883: The header field/key must also be trimmed of whitespace

source/sigv4.c
```diff
@@ -793,6 +854,230 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+            }
+            else
+            {
+                if( ( ( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) && ( pHeaders[ i ] == '\n' ) ) || ( !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) && ( pHeaders[ i ] == '\r' ) && ( ( i + 1 ) < headersLen ) && ( pHeaders[ i + 1 ] == '\n' ) ) )
+                {
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].value.pData = start;
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].value.dataLen = ( end - start );
+                    start = end + 1U;
+                    fieldFlag = 1;
+                    noOfHeaders + 1U;
```
yourslab, line 1056: This doesn't actually increment `noOfHeaders`

source/sigv4.c
gshvang: This was done for testing things locally. While integration we need to decide whether we need to call it separately or as a part of this function based on that we can add the new line character.

source/sigv4.c
gshvang: The following pseudocode describes how to construct the canonical list of headers:
```
CanonicalHeaders =
CanonicalHeadersEntry0 + CanonicalHeadersEntry1 + ... + CanonicalHeadersEntryN
CanonicalHeadersEntry =
Lowercase(HeaderName) + ':' + Trimall(HeaderValue) + '\n'
```
As per the AWS doc it does not mention about trimming header key.

yourslab: To create the canonical headers list, **convert all header names to lowercase and remove leading spaces and trailing spaces**. Convert sequential spaces in the header value to a single space.

source/include/sigv4.h
```diff
@@ -150,7 +150,16 @@ typedef enum SigV4Status
      * Functions that may return this value:
      * - #SigV4_AwsIotDateToIso8601
      */
-    SigV4ISOFormattingError
+    SigV4ISOFormattingError,
+
+    /**
+     * @brief The maximum number of header parameters was exceeded while parsing
+     * the http header string input parameter.
+     *
```
aggarw13, line 158: If there is an upper limit on the number of header parameters supported for parsing by the library specified in a macro, it can be mentioned here.

source/sigv4.c
```diff
@@ -246,6 +246,67 @@ static SigV4Status_t verifySigV4Parameters( const SigV4Parameters_t * pParams );
...
+ * @param[out] canonicalRequest Struct to maintain intermediary buffer
+ * and state of canonicalization.
+ */
+static SigV4Status_t generateCanonicalHeaders( const char * pHeaders,
+                                               size_t headersLen,
+                                               uint32_t flags,
+                                               CanonicalContext_t * canonicalRequest );
+
+/**
+ * @brief Append Signed Headers to the string which needs to be signed.
```
aggarw13, line 265: "Signed Headers" are not part of the "String to sign" according to the documentation: https://docs.aws.amazon.com/general/latest/gr/sigv4-create-string-to-sign.html

Does this statement need to be updated?

source/sigv4.c
```diff
@@ -793,6 +854,236 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
     }
 
 /*-----------------------------------------------------------*/
+    static SigV4Status_t writeSignedHeaderToString( size_t headerIndex,
```
aggarw13, line 857: As this function is only writing the "Signed Header" to the Canonical Request Buffer, it would be better to name is appropriately.
```suggestion
    static SigV4Status_t writeSignedHeaderToCanonicalRequest( size_t headerIndex,
```

source/sigv4.c
aggarw13: Is this function writing "SignedHeaders" or "CanonicalHeaders" to the CanonicalRequest? The following is the formula of the CanonicalRequest where these 2 are distinct elements of the equation:
```
CanonicalRequest =
  HTTPRequestMethod + '\n' +
  CanonicalURI + '\n' +
  CanonicalQueryString + '\n' +
  CanonicalHeaders + '\n' +
  SignedHeaders + '\n' +
  HexEncode(Hash(RequestPayload))
  ```

File: source/sigv4.c [hunk 25]
```diff
@@ -793,6 +854,236 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                }
+                else
+                {
+                    *pBufLoc = canonicalRequest->pHeadersLoc[ headerNum ].key.pData[ i ];
+                }
+
+                pBufLoc++;
+            }
+
+            *pBufLoc = ';';
```
aggarw13, line 893: How does this function take care of not adding an extra `;` after the last header?

aggarw13: I see that another function that calls this function takes care of omitting the extra `;`, please add a comment about that

source/sigv4.c
```diff
@@ -793,6 +854,236 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        if( sigV4Status == SigV4Success )
+        {
+            canonicalRequest->pBufCur--;
+            *canonicalRequest->pBufCur = '\n';
+        }
+
+        return sigV4Status;
+    }
+
+    static SigV4Status_t writeCanonicalHeaderToString( size_t headerIndex,
```
aggarw13, line 933: As this is generating the "CanonicalRequest", it should be named appropriately to not be confused with the string related to "String To Sign"
```suggestion
    static SigV4Status_t writeCanonicalHeaderToCanonicalRequest( size_t headerIndex,
```

source/sigv4.c
```diff
@@ -793,6 +854,236 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+            if( sigV4Status != SigV4Success )
+            {
+                break;
+            }
+        }
+
+        return sigV4Status;
+    }
+
+    static SigV4Status_t generateCanonicalHeaders( const char * pHeaders,
```
aggarw13, line 1012: This function is creating "CanonicalRequest" instead of "CanonicalHeaders"...
```suggestion
    static SigV4Status_t generateCanonicalRequest( const char * pHeaders,
```

source/sigv4.c
```diff
@@ -793,6 +854,236 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+
+        assert( canonicalRequest != NULL );
+        assert( canonicalRequest->pBufCur != NULL );
+
+        pBufLoc = canonicalRequest->pBufCur;
+        buffRemaining = canonicalRequest->bufRemaining;
+
+        keyLen = canonicalRequest->pHeadersLoc[ headerNum ].key.dataLen;
+
+        if( keyLen + 1 > buffRemaining )
```
aggarw13, line 873: Please add a comment here about `+1` being added for `;`. Also, please add similar comments in other functions that perform length check with `+ x` arithment

source/sigv4.c
```diff
@@ -793,6 +854,236 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+            value = canonicalRequest->pHeadersLoc[ headerNum ].value.pData;
+            trimValueLen = 0;
+
+            for( i = 0; i < valLen; i++ )
+            {
+                if( !( isspace( value[ i ] ) && ( i + 1 <= valLen ) && ( ( i + 1 == valLen ) || isspace( value[ i + 1 ] ) || ( trimValueLen == 0 ) ) ) )
+                {
+                    *pBufLoc = canonicalRequest->pHeadersLoc[ headerNum ].value.pData[ i ];
+                    pBufLoc++;
+                    trimValueLen + 1U;
```
aggarw13, line 973: What is this statement doing? It will not have any effect on the value of `trimValueLen`

source/sigv4.c
aggarw13: Please update this expression into a more readable form by creating private functions that allow the expression to look like:
`if ( ! isLeadingSpace( value, i ) && !  isTrailingSpace( value, i ) && ! isFirsInSequentialSpaces( value, i ) )`

source/sigv4.c
```diff
@@ -793,6 +854,236 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+
+        pBufLoc = canonicalRequest->pBufCur;
+        buffRemaining = canonicalRequest->bufRemaining;
+
+        keyLen = canonicalRequest->pHeadersLoc[ headerNum ].key.dataLen;
+        valLen = canonicalRequest->pHeadersLoc[ headerNum ].value.dataLen;
+
+        if( keyLen + valLen + 2 > buffRemaining )
+        {
+            sigV4Status = SigV4InsufficientMemory;
```
aggarw13, line 952: If the Header Value contains spaces that will be trimmed, then this logic can result in a false failure. Thus, it would be better to check for buffer space while actually copying the trimmed header value byte by byte. 
The logic I suggested in [this comment](https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/pull/28#discussion_r682165080) cane be updated to also check for buffer space availability before copying the header value byte into the canonical request buffer:

```
if ( ! isLeadingSpace( value, valueLen, i ) && ! isTrailingSpace( value, valueLen, i ) && ! isFirsInSequentialSpaces( value, valueLen, i ) )
{ 
    if( bufferRemaining < 2 )
    {
        sigV4Status = SigV4InsufficientMemory; 
        break:
    }
    else
    {
           *pBufLoc = canonicalRequest->pHeadersLoc[ headerNum ].value.pData[ i ];
             pBufLoc++;
             buffRemaining--;
     }
}
```

source/sigv4.c
muneebahmed10: > I think this piece of code just removes all spaces from the string and needs more testing

The condition inside the `if` is basically:
```
!( isCurrentCharSpace && ( true ) && ( ( isLastChar ) || isNextCharSpace || ( isLeadingCharacter ) ) )
```
which further boils down to:
```
!( isCurrentCharSpace && ( isLastChar || isNextCharSpace || isLeadingCharacter ) )
```
which looks like it would correctly trim spaces, it's just confusing to think about and could probably be replaced with a function to be more readable

muneebahmed10: Copying my suggestion from slack:
```C
static bool isTrimmableSpace( const char * value, size_t index, size_t valLen, size_t trimmedLength ) {
    bool ret = false;

    /* Only trim spaces. */
    if( isspace( value[ index ] ) )
    {
        /* The last character is a trailing space. */
        if( ( index + 1 ) == valLen )
        {
            ret = true;
        }
        /* Trim if the next character is also a space. */
        else if( isspace( value[ index + 1 ] ) )
        {
            ret = true;
        }
        /* It is a leading space if no characters have been written yet. */
        else if( trimmedLength == 0U )
        {
            ret = true;
        }
    }
    return ret;
}
```

source/sigv4.c
gshvang: This is adding Signed Headers to the canonical Request. Although the name mentions about adding Signed headers to the cancel request. I will rename it again so as to make it more comprehnsible.

source/sigv4.c
gshvang: Signed Headers are part of canonical request which is turn are hashed and added to string we need to sign to get the auth signature.

yourslab: It's more appropriate to say that it is appended to the canonical request

source/sigv4.c
```diff
@@ -793,6 +869,330 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        /* Sorting headers based on keys. */
+        if( ( sigV4Status == SigV4Success ) && !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) )
+        {
+            qsort( canonicalRequest->pHeadersLoc, noOfHeaders, sizeof( SigV4KeyValuePair_t ), cmpField );
+        }
+
+        if( ( sigV4Status == SigV4Success ) && !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) )
+        {
+            sigV4Status = appendCanonicalizedHeaders( noOfHeaders, canonicalRequest );
+        }
```
muneebahmed10, line 1173: It doesn't look like `sigV4Status` is changed in the first block so they can be combined? Also, if the `SIGV4_HTTP_PATH_IS_CANONICAL_FLAG` is set then will the canonical headers never be appended?
```suggestion
        if( ( sigV4Status == SigV4Success ) && !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) )
        {
            qsort( canonicalRequest->pHeadersLoc, noOfHeaders, sizeof( SigV4KeyValuePair_t ), cmpField );
            sigV4Status = appendCanonicalizedHeaders( noOfHeaders, canonicalRequest );
        }
```

source/sigv4.c
```diff
@@ -793,6 +869,330 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        assert( canonicalRequest != NULL );
+        assert( canonicalRequest->pBufCur != NULL );
+
+        start = end = pHeaders;
+
+        for( i = 0; i < headersLen; i++ )
+        {
+            if( noOfHeaders == SIGV4_MAX_HTTP_HEADER_COUNT )
+            {
+                sigV4Status = SigV4MaxHeaderPairCountExceeded;
```
muneebahmed10, line 1127: You should `break` at this point too, right?

yourslab (commented):
Pls bring down the complexity when you can as we are almost at <= 8 for all functions 🎉

source/sigv4.c
```diff
@@ -793,6 +869,330 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+            {
+                if( ( !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) && ( pHeaders[ i ] == '\r' ) && ( ( i + 1 ) < headersLen ) && ( pHeaders[ i + 1 ] == '\n' ) ) )
+                {
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].value.pData = start;
+                    canonicalRequest->pHeadersLoc[ noOfHeaders ].value.dataLen = ( end - start );
+                    start = end + 2U;
+                    fieldFlag = 1;
+                    noOfHeaders++;
+                }
+                else if( ( ( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) && ( pHeaders[ i ] == '\n' ) ) )
```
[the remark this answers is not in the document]
gshvang: If the `SIGV4_HTTP_PATH_IS_CANONICAL_FLAG `is not set , the headers are already canonicalized. And if they are not canonicalized then they will surely be having '\r\n' at the end of each header key-value pair as per mentioned in coreHTTP Library.

lexicon.txt
```diff
@@ -1,6 +1,7 @@
 
 accesskeyid
 accesskeylen
+accommodate
```
muneebahmed10, line 4: Why does this need to be added to the lexicon when it's already a word?

lexicon.txt
```diff
@@ -68,6 +71,7 @@ mon
 monthsperday
 noninfringement
 nullterminate
+occurrences
```
muneebahmed10, line 74: Same here

source/sigv4.c
```diff
@@ -246,6 +246,107 @@ static SigV4Status_t verifySigV4Parameters( const SigV4Parameters_t * pParams );
...
+/**
+ * @brief Helper function to check whether the current character is a
+ * space that is either trailing, leading, or having multiple occurrences
+ * between characters in a string.
+ *
+ * @param[in] value String to be trimmed.
+ * @param[in] index Index of current character.
+ * @param[in] valLen Length of the string.
+ * @param[in] trimmedLength length of string after trimming.
+ * and state of canonicalization.
```
muneebahmed10, line 341:
```suggestion
 * @param[in] trimmedLength Current length of trimmed string.
```

source/sigv4.c
```diff
@@ -794,6 +895,336 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        if( sigV4Status == SigV4Success )
+        {
+            /* Replacing the last ';' with '\n' as last header does need to have ';'. */
+            canonicalRequest->pBufCur--;
+            *canonicalRequest->pBufCur = '\n';
+            canonicalRequest->pBufCur++;
+        }
+
+        return sigV4Status;
+    }
```
muneebahmed10, line 1022: Please use `/*-----------------------------------------------------------*/` to separate each function implementation
```suggestion
    }

/*-----------------------------------------------------------*/
```

source/sigv4.c
```diff
@@ -794,6 +895,336 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                break;
+            }
+        }
+
+        if( sigV4Status == SigV4Success )
+        {
+            /* Replacing the last ';' with '\n' as last header does need to have ';'. */
+            canonicalRequest->pBufCur--;
+            *canonicalRequest->pBufCur = '\n';
+            canonicalRequest->pBufCur++;
```
muneebahmed10, line 1018:
```suggestion
            /* Replace the last ';' with '\n' as the last header does need it. */
            *( canonicalRequest->pBufCur - 1 ) = '\n';
```

source/sigv4.c
```diff
@@ -794,6 +895,336 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+            assert( ( canonicalRequest->pHeadersLoc[ noOfHeaders ].key.pData ) != NULL );
+            sigV4Status = writeSignedHeaderToCanonicalRequest( noOfHeaders, flags, canonicalRequest );
+
+            if( sigV4Status != SigV4Success )
+            {
+                break;
+            }
+        }
+
+        if( sigV4Status == SigV4Success )
```
muneebahmed10, line 1013: Is it at all possible that this function was invoked with zero headers?
```suggestion
        if( ( sigV4Status == SigV4Success ) && ( headerCount > 0U ) )
```

source/sigv4.c
```diff
@@ -794,6 +895,336 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+
+            end++;
+        }
+
+        /* Sorting headers based on keys. */
+        if( ( sigV4Status == SigV4Success ) && !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) )
+        {
+            qsort( canonicalRequest->pHeadersLoc, noOfHeaders, sizeof( SigV4KeyValuePair_t ), cmpField );
+
+            sigV4Status = appendCanonicalizedHeaders( noOfHeaders, canonicalRequest );
```
muneebahmed10, line 1203: Why don't the headers need to be written into the buffer when they are already canonicalized?

source/sigv4.c
```diff
@@ -854,12 +894,44 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                ret = true;
+            }
+            /* It is a leading space if no characters have been written yet. */
+            else if( trimmedLength == 0U )
+            {
+                ret = true;
+            }
+        }
+
+        return ret;
```
aggarw13, line 925: Alternative suggestion to make the code very explicit about the different types of spaces that are trimmed.
```suggestion

    bool isCurrCharSpace = isSpace( value[index] ); 
    bool isLeadingSpace = isCurrCharSpace && ( trimmedLength == 0 );
    bool isTrailingSpace = isCurrCharSpace && ( ( index + 1 ) == valLen ); 
    bool isNotLastInSequentialSpaces = isCurrCharSpace && ( (index + 1) < valLen) && (isSpace( value[ index + 1 ] ) );

    return ( isLeadingSpace || isTrailingSpace || isNotLastInSequentialSpaces ) ;
```

muneebahmed10: Disagree with this since the triple condition for the last bool is hard to read, and the original comments are clear enough as to what's going on. There's no short circuit evaluation here so all condition are evaluated, unlike the `else if` case, where it only evaluates as many conditions as is needed.

source/sigv4.c
```diff
@@ -292,20 +306,46 @@ static SigV4Status_t appendCanonicalizedHeaders( size_t headerCount,
  * in the canonical form.
  * @param[in,out] canonicalRequest Struct to maintain intermediary buffer
  * and state of canonicalization.
+ *
+ * @return Following statuses will be returned by the function:
+ * #SigV4Success if headers are successfully added to the canonical request.
```
aggarw13, line 311:
```suggestion
 * #SigV4Success if the signed header is successfully added to the canonical request.
```

source/sigv4.c
```diff
@@ -292,20 +306,46 @@ static SigV4Status_t appendCanonicalizedHeaders( size_t headerCount,
...
+ * #SigV4InsufficientMemory if buffer remaining cannot accommodate all the provided headers.
```
aggarw13, line 312:
```suggestion
 * #SigV4InsufficientMemory if canonical request buffer cannot accommodate the header.
```

source/sigv4.c
```diff
@@ -292,20 +306,46 @@ static SigV4Status_t appendCanonicalizedHeaders( size_t headerCount,
...
+ */
+static SigV4Status_t writeCanonicalHeaderToCanonicalRequest( size_t headerIndex,
+                                                             CanonicalContext_t * canonicalRequest );
+
+/**
+ * @brief Helper function to check whether the current character is a
+ * space that is either trailing, leading, or having multiple occurrences
+ * between characters in a string.
```
aggarw13, line 335:
```suggestion
 * @brief Helper function to determine whether a header value character represents a space 
 * that can be trimmed when for creating "CanonicalHeaders". 
 * All leading and trailing spaces in the header value need to be trimmed. Also, sequential spaces in the 
 * header value need to be trimmed to a single space."
 *
 * Example of modifying header field for CanonicalHeaders: 
 * Actual header pair:
 * My-Header2:    "a   b   c"  \n
 * 
 * Modified header pair for "CanonicalHeaders":
 * my-header2:"a b c"\n
```

source/sigv4.c
```diff
@@ -292,20 +306,46 @@ static SigV4Status_t appendCanonicalizedHeaders( size_t headerCount,
...
+ *
+ * @param[in] value String to be trimmed.
```
aggarw13, line 337:
```suggestion
 * @param[in] value Header value to be trimmed.
```

gshvang: It can be header value or key so should not we keep it more generalized.

source/sigv4.c
aggarw13: Please make similar doc update to `writeCanonicalHeaderToCanonicalRequest`

source/sigv4.c
gshvang: If header count is zero it will not even go inside the loop. Also as mentioned in the AWS docs there will be  1 header for sure (For HTTP/1.1 requests, you must include the host header at a minimum.)

source/sigv4.c
gshvang: If they are already canonicalized we are just copying them using different function into the buffer. `writeLineToCanonicalRequest`which is part of Oscar's PR  is used for it.

source/sigv4.c
aggarw13: I generally think it carries better maintenance value in writing code in an expressive style but maybe in this case the advantage of short-circuit evaluation of `if-else` statements may be important as this function is called for all characters of header pairs.  

@gshvang, please do add the assert check though to the function

gshvang: Yeah I added the assert check.

source/sigv4.c
```diff
@@ -794,6 +895,336 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        }
+        else
+        {
+            *pBufLoc = ';';
+            pBufLoc++;
+            /* Calculate the remaining buffer. 1 is added for ';' character. */
+            buffRemaining -= 1;
+            canonicalRequest->pBufCur = pBufLoc;
+            canonicalRequest->bufRemaining = buffRemaining;
+        }
```
aggarw13, line 995:
```suggestion
To avoid duplication of buffer copying code as well as add some helpful comments.
```suggestion
        bool shouldCopyChar = false;
        char charToCopy;
        size_t numOfCopiedBytes = 0;
        for( i = 0; i < keyLen; i++ )
        {
            shouldCopyChar = false;
            if( !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) )
            {
                if( !( isTrimmableSpace( headerKey, i, keyLen, numOfCopiedBytes ) ) )
                {
                    charToCopy = tolower( headerKey[ i ] );
                    shouldCopyChar = true;
                }
            }
            else
            {
                charToCopy= canonicalRequest->pHeadersLoc[ headerIndex ].key.pData[ i ];
                shouldCopyChar = true;
            }
            
            /* Remaining buffer space should at least accommodate the character to copy and the trailing ";" */
            if( shouldCopyChar &&  ( buffRemaining <= 1 ) )
            {
                sigV4Status = SigV4InsufficientMemory;
            }
            else if( shouldCopyChar && ( buffRemaining > 1 ) ) 
            {
                *pBufLoc = charToCopy;
                pBufLoc++;
                buffRemaining =-1;
                numOfCopiedBytes;
            }
        }
        
        /* Space for ending ";" was accounted for while copying header field data to
          * canonical request buffer. */
        if( sigV4Status == SigV4Success )
        {
            assert( buffRemaining >= 1 );
            *pBufLoc = ';';
            pBufLoc++;
            /* Calculate the remaining buffer. 1 is added for ';' character. */
            buffRemaining -= 1;
            canonicalRequest->pBufCur = pBufLoc;
            canonicalRequest->bufRemaining = buffRemaining;
        }
```

source/sigv4.c
aggarw13: Would be better to call it "Header value or key string"

source/sigv4.c
```diff
@@ -794,6 +899,339 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+
+        keyLen = canonicalRequest->pHeadersLoc[ headerIndex ].key.dataLen;
+
+        headerKey = canonicalRequest->pHeadersLoc[ headerIndex ].key.pData;
+
+        for( i = 0; i < keyLen; i++ )
+        {
+            /* If the header field is not in canonical form already, we need to check
+             * whether this character represents a trimmable space. */
+            if( !( flags & SIGV4_HTTP_PATH_IS_CANONICAL_FLAG ) &&
```
muneebahmed10, line 959: Since this is for headers, why does it use `SIGV4_HTTP_PATH_IS_CANONICAL_FLAG` instead of `SIGV4_HTTP_HEADERS_ARE_CANONICAL_FLAG`?

source/sigv4.c
muneebahmed10: Would `assert( headerCount > 0 );` be worth adding then?

source/sigv4.c
aggarw13:
```suggestion
 * @brief Append "Signed Headers" to the Canonical Request buffer.
```

source/sigv4.c
```diff
@@ -246,6 +246,111 @@ static SigV4Status_t verifySigV4Parameters( const SigV4Parameters_t * pParams );
...
+ * @return Following statuses will be returned by the function:
+ * #SigV4Success if headers are successfully added to the canonical request.
+ * #SigV4InsufficientMemory if canonical request buffer cannot accommodate the header.
+ */
+static SigV4Status_t appendSignedHeaders( size_t headerCount,
+                                          uint32_t flags,
+                                          CanonicalContext_t * canonicalRequest );
+
+/**
+ * @brief Canonicalize headers and append it to the string which needs to be signed.
```
aggarw13, line 288:
```suggestion
 * @brief Canonicalize headers and append it to the Canonical Request buffer.
```

source/sigv4.c
```diff
@@ -794,6 +899,332 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+    }
+
+/*-----------------------------------------------------------*/
+
+    static SigV4Status_t writeSignedHeaderToCanonicalRequest( size_t headerIndex,
+                                                              uint32_t flags,
+                                                              CanonicalContext_t * canonicalRequest )
+    {
+        char * pBufLoc;
+        size_t buffRemaining, keyLen = 0, i = 0, trimKeyLen = 0;
```
aggarw13, line 941:
```suggestion
        size_t buffRemaining, keyLen = 0, i = 0, curNumOfCopiedBytes = 0;
```

source/sigv4.c
```diff
@@ -794,6 +899,332 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        return sigV4Status;
+    }
+
+/*-----------------------------------------------------------*/
+
+    static SigV4Status_t appendSignedHeaders( size_t headerCount,
+                                              uint32_t flags,
+                                              CanonicalContext_t * canonicalRequest )
+    {
+        size_t noOfHeaders = 0, keyLen = 0, i = 0;
```
aggarw13, line 999: `keyLen` and `i` are unused
```suggestion
        size_t noOfHeaders = 0;
```

source/sigv4.c
```diff
@@ -794,6 +899,332 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                *pBufLoc = '\n';
+                pBufLoc++;
+
+                /* Calculate the remaining buffer. 1 is added for '\n' character. */
+                buffRemaining -= 1;
+
+                canonicalRequest->pBufCur = pBufLoc;
+                canonicalRequest->bufRemaining = buffRemaining;
+            }
+        }
```
aggarw13, line 1112: The logic of copying header key and value is repeated thrice across `writeCanonicalHeaderToCanonicalRequest` and `writeSignedHeaderToCanonicalRequest` functions. I instead suggesting creating a single function that performs this common logic of copying the header string in a trimmed fashion. With this common function, the `writeCanonicalHeaderToCanonicalRequest` and `writeSignedHeaderToCanonicalRequest` shouldn't be required and the common function can be called directly in `appendSignedHeaders` and `appendCanonicalizedHeaders` functions respectively.


```
SigV4Status_t copyHeaderStringToCanonicalBuffer( const char * pData,
                                                 size_t dataLen,
                                                 char * pCanonicalBuffer,
                                                 size_t bufferSize,
                                                 char separator,
                                                 size_t * pCopiedStringLen )
{
    Sigv4Status_t status = SigV4Success;
    size_t index = 0;
    size_t numOfBytesCopied = 0;
    size_t buffRemaining = bufferSize;
    char * pCurrBufLoc = pCanonicalBuffer;

    assert( ( pData != NULL ) && ( dataLen > 0 ) );
    assert( pCanonicalBuffer != NULL );
    assert( pCopiedStringLen != NULL );

    for( index = 0; index < dataLen; index++ )
    {
        if( isTrimmableSpace( pData, i, dataLen, numOfBytesCopied ) )
        {
            /* Cannot copy trimmable space into canonical request buffer. */
        }
        /* Remaining buffer space should at least accommodate the character to copy and the trailing ":" */
        else if( buffRemaining <= 1 )
        {
            status = SigV4InsufficientMemory;
            break;
        }
        else
        {
            *pCurrBufLoc = tolower( pData[ index ] );
            pCurrBufLoc++;
            numOfBytesCopied++;
        }
    }

     /* Add the ending separating character passed to the function.
      * Note: Space for the separator character is accounted for while copying
      * header field data to canonical request buffer. */
    if( status == SigV4Success )
    {
        assert( buffRemaining >= 1 );
        *pCurrBufLoc = separator;

        *pCopiedStringLen = numOfBytesCopied;
    }


    return status;
}
```

source/sigv4.c
```diff
@@ -794,6 +899,332 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+
+        return sigV4Status;
+    }
+
+/*-----------------------------------------------------------*/
+
+    static SigV4Status_t appendCanonicalizedHeaders( size_t headerCount,
+                                                     CanonicalContext_t * canonicalRequest )
+    {
+        size_t noOfHeaders = 0;
```
aggarw13, line 1122: `headerIndex` is more suitable of a name for the iteration over the list of header pairs
```suggestion
        size_t headerIndex= 0;
```

source/sigv4.c
```diff
@@ -794,6 +899,332 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        const char * start;
+        const char * end;
+        size_t keyFlag = 1, noOfHeaders = 0, i = 0;
+        SigV4Status_t sigV4Status = SigV4Success;
+
+        assert( pHeaders != NULL );
+        assert( canonicalRequest != NULL );
+        assert( canonicalRequest->pBufCur != NULL );
+
+        start = end = pHeaders;
```
aggarw13, line 1159: I haven't seen us using double assignment before. Is it acceptable within our coding style?

source/sigv4.c
```diff
@@ -794,6 +899,332 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+/*-----------------------------------------------------------*/
+
+    static SigV4Status_t appendAllHeadersToCanonicalRequest( const char * pHeaders,
+                                                             size_t headersLen,
+                                                             uint32_t flags,
+                                                             CanonicalContext_t * canonicalRequest )
+    {
+        const char * start;
+        const char * end;
+        size_t keyFlag = 1, noOfHeaders = 0, i = 0;
```
aggarw13, line 1152: If `keyFlag` just carries binary value, this can be just of `bool` type.

source/sigv4.c
```diff
@@ -794,6 +899,332 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+            else if( ( keyFlag == 0 ) && ( ( flags & SIGV4_HTTP_HEADERS_ARE_CANONICAL_FLAG ) && ( pHeaders[ i ] == '\n' ) ) )
+            {
+                canonicalRequest->pHeadersLoc[ noOfHeaders ].value.pData = start;
+                canonicalRequest->pHeadersLoc[ noOfHeaders ].value.dataLen = ( end - start );
+                start = end + 1U;
+                keyFlag = 1;
+                noOfHeaders++;
+            }
+
+            end++;
```
aggarw13, line 1193: `end` is carrying redundant information as `i`, so, `end` can be removed

source/sigv4.c
```diff
@@ -794,6 +899,332 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        }
```
aggarw13, line 1194: To reduce the complexity of this function, the logic of parsing the header key-value pairs can be culled out into a separate function. Here's my suggestion of the logic with some variable naming improvements.

```
SigV4Status_t parseHeaderKeyValueEntries( const char * pHeadersData,
                                          size_t headersDataLen,
                                          uint32_t flags,
                                          CanonicalContext_t * pCanonicalRequest )
{
    assert( pHeadersData != NULL );
    assertheadersDataLen > 0 );
    assert( pCanonicalRequest != NULL );

    size_t index = 0;
    const char * pKeyOrValStartLoc = pHeadersData;
    const char * pCurrLoc = pHeadersData;
    bool keyFound = false;

    for( index = 0; index < assertheadersDataLen; i++ )
    {
        pCurrLoc = &pHeadersData[index];
        if( noOfHeaders == SIGV4_MAX_HTTP_HEADER_COUNT )
        {
            sigV4Status = SigV4MaxHeaderPairCountExceeded;
            break;
        }
        /* Look for key part of an header field entry. */
        else if( ( keyFound == false ) && ( pHeaders[ index ] == ':' ) )
        {
            pCanonicalRequest->pHeadersLoc[ noOfHeaders ].key.pData = pKeyOrValStartLoc;
            pCanonicalRequest->pHeadersLoc[ noOfHeaders ].key.dataLen = ( pCurrLoc - pKeyOrValStartLoc );
            pKeyOrValStartLoc = pCurrLoc + 1U;
            keyFound = false;
        }
        /* Look for header value part of a header field entry for both canonicalized and non-canonicalized forms. */
        else if( keyFound == true)
        {
            /* Non-canonicalized headers will have header values ending with "\r\n". */
            if( !( flags & SIGV4_HTTP_HEADERS_ARE_CANONICAL_FLAG ) && ( ( index + 1 ) < headersLen ) 
                    && (0 == strncmp(pCurrLoc, "\r\n", strlen("\r\n"))) ) 
            {
                pCanonicalRequest->pHeadersLoc[ noOfHeaders ].value.pData = pKeyOrValStartLoc;
                pCanonicalRequest->pHeadersLoc[ noOfHeaders ].value.dataLen = ( pCurrLoc - pKeyOrValStartLoc );
                /* Set starting location of the next header key string after the "\r\n". */
                pKeyOrValStartLoc = pCurrLoc + 2U;
                keyFound = false;
                noOfHeaders++;
            }
            /* Canonicalized headers will have header values ending just with "\n". */
            else if( !( flags & SIGV4_HTTP_HEADERS_ARE_CANONICAL_FLAG ) && ( pHeaders[ index ] == '\n' ) )
            {
                pCanonicalRequest->pHeadersLoc[ noOfHeaders ].value.pData = pKeyOrValStartLoc;
                pCanonicalRequest->pHeadersLoc[ noOfHeaders ].value.dataLen = ( pCurrLoc - pKeyOrValStartLoc );
                /* Set starting location of the next header key string after the "\n". */
                pKeyOrValStartLoc = pCurrLoc + 1U;
                keyFound = false;
                noOfHeaders++;
            }
            else
            {
                sigV4Status = Sigv4BadParameter;
                break;
            }
        }
    }

    return sigV4Status;
}
```

gshvang: `pCurrLoc = &pHeadersData[index];` . This line will update the currentLoc pointer every time but we want to update it only when either we reach the end of key or value.

gshvang:
```
else
            {
                sigV4Status = Sigv4BadParameter;
                break;
            }
```

This part necessarily need not be always Invalid Parameter. It can be any other character in the value too.

source/sigv4.c
```diff
@@ -296,38 +299,54 @@ static SigV4Status_t appendSignedHeaders( size_t headerCount,
...
  */
-static SigV4Status_t writeSignedHeaderToCanonicalRequest( size_t headerIndex,
-                                                          uint32_t flags,
-                                                          CanonicalContext_t * canonicalRequest );
+SigV4Status_t parseHeaderKeyValueEntries( const char * pHeaders,
+                                          size_t headersDataLen,
+                                          uint32_t flags,
+                                          size_t * headerCount,
+                                          ,
+                                          CanonicalContext_t * canonicalRequest );
```
aggarw13, line 328:
```suggestion
                                          size_t * headerCount,
                                          CanonicalContext_t * canonicalRequest );
```

source/sigv4.c
```diff
@@ -296,38 +299,54 @@ static SigV4Status_t appendSignedHeaders( size_t headerCount,
...
+ * @param[in] dataLen Length of Header Key or value.
+ * @param[in] flags Flag to indicate if headers are already
+ * in the canonical form.
+ * @param[in] separator Character separating the multiple key-value pairs or key and values.
  * @param[in,out] canonicalRequest Struct to maintain intermediary buffer
  * and state of canonicalization.
  *
  * @return Following statuses will be returned by the function:
- * #SigV4Success if the canonical headers are successfully added to the canonical request.
+ * #SigV4Success if the signed headers are successfully added to the canonical request.
```
aggarw13, line 342:
```suggestion
 * #SigV4Success if the header key or value is successfully added to the canonical request.
```

source/sigv4.c
```diff
@@ -246,6 +246,130 @@ static SigV4Status_t verifySigV4Parameters( const SigV4Parameters_t * pParams );
 static void hexEncode( SigV4String_t * pInputStr,
                        SigV4String_t * pHexOutput );
 
+/**
+ * @brief Extract each header key and value from passed pHeaders string.
```
aggarw13, line 250:
```suggestion
 * @brief Extract all header key-value pairs from the passed headers data and add them
 * to the canonical request.
```

source/sigv4.c
```diff
@@ -246,6 +246,130 @@ static SigV4Status_t verifySigV4Parameters( const SigV4Parameters_t * pParams );
...
+ * @param[in] headersDataLen Length of HTTP headers to parse.
+ * @param[in] flags Flag to indicate if headers are already
+ * in the canonical form.
+ * @param[out] headerCount Count of key-value pairs parsed from pData.
+ * @param[out] canonicalRequest Struct to maintain intermediary buffer
+ * and state of canonicalization.
+ *
+ * @return Following statuses will be returned by the function:
+ * #SigV4Success if headers are successfully added to the canonical request.
+ * #SigV4InsufficientMemory if canonical request buffer cannot accommodate the header.
```
aggarw13, line 318: Is this return code valid for the function?

source/sigv4.c
```diff
@@ -246,6 +246,130 @@ static SigV4Status_t verifySigV4Parameters( const SigV4Parameters_t * pParams );
...
+ * #SigV4InvalidParameter if HTTP headers are invalid.
+ * #SigV4MaxHeaderPairCountExceeded if number of headers that needs to be canonicalized
+ * exceed the SIGV4_MAX_HTTP_HEADER_COUNT macro defined in the config file.
```
aggarw13, line 321:
```suggestion
 * #SigV4MaxHeaderPairCountExceeded if number of key-value entries in the headers data
 * exceeds the SIGV4_MAX_HTTP_HEADER_COUNT macro defined in the config file.
```

source/sigv4.c
```diff
@@ -246,6 +246,130 @@ static SigV4Status_t verifySigV4Parameters( const SigV4Parameters_t * pParams );
...
+ */
+SigV4Status_t parseHeaderKeyValueEntries( const char * pHeaders,
+                                          size_t headersDataLen,
+                                          uint32_t flags,
+                                          size_t * headerCount,
+                                          ,
+                                          CanonicalContext_t * canonicalRequest );
+
+/**
+ * @brief Copy headers key or header value to the Canonical Request buffer.
```
aggarw13, line 331:
```suggestion
 * @brief Copy header key or header value to the Canonical Request buffer.
```

source/sigv4.c
```diff
@@ -794,6 +918,308 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        for( index = 0; index < dataLen; index++ )
+        {
+            /* If the header field is not in canonical form already, we need to check
+             * whether this character represents a trimmable space. */
+            if( !( flags & SIGV4_HTTP_HEADERS_ARE_CANONICAL_FLAG ) &&
+                isTrimmableSpace( pData, index, dataLen, numOfBytesCopied ) )
+            {
+                /* Cannot copy trimmable space into canonical request buffer. */
+            }
+            /* Remaining buffer space should at least accommodate the character to copy and the trailing ":" */
```
aggarw13, line 983:
```suggestion
            /* Remaining buffer space should at least accommodate the character to copy and the trailing 
             * separator character*/
```

source/sigv4.c
```diff
@@ -794,6 +918,308 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                status = SigV4InsufficientMemory;
+                break;
+            }
+            else
+            {
+                /* Lowercase header key only. '\n' character marks the ending of value. */
+                if( separator == '\n' )
+                {
+                    *pCurrBufLoc = ( pData[ index ] );
+                }
```
aggarw13, line 995: What is the reason for not using `tolower` for `\n` as the separator character? The comment does not give me a clear reasoning for the special handling

source/sigv4.c
```diff
@@ -794,6 +918,308 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+
+        /* Add the ending separating character passed to the function.
+         * Note: Space for the separator character is accounted for while copying
+         * header field data to canonical request buffer. */
+        if( status == SigV4Success )
+        {
+            assert( buffRemaining >= 1 );
+            *pCurrBufLoc = separator;
+            pCurrBufLoc++;
+            numOfBytesCopied++;
```
aggarw13, line 1015:
```suggestion
```

source/sigv4.c
```diff
@@ -794,6 +918,308 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        assert( headerCount > 0 );
+
+        for( headerIndex = 0; headerIndex < headerCount; headerIndex++ )
+        {
+            assert( ( canonicalRequest->pHeadersLoc[ headerIndex ].key.pData ) != NULL );
+            keyLen = canonicalRequest->pHeadersLoc[ headerIndex ].key.dataLen;
+
+            headerKey = canonicalRequest->pHeadersLoc[ headerIndex ].key.pData;
+
+            sigV4Status = copyHeaderStringToCanonicalBuffer( headerKey, keyLen, flags, ';', canonicalRequest );
```
aggarw13, line 1044: Would be good to add a comment here that `;` is used to separate the "Signed Headers"

source/sigv4.c
```diff
@@ -794,6 +918,308 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+        const char * headerKey;
+        SigV4Status_t sigV4Status = SigV4Success;
+
+        assert( canonicalRequest != NULL );
+        assert( canonicalRequest->pBufCur != NULL );
+        assert( headerCount > 0 );
+
+        for( headerIndex = 0; headerIndex < headerCount; headerIndex++ )
+        {
+            assert( canonicalRequest->pHeadersLoc[ headerIndex ].key.pData != NULL );
```
aggarw13, line 1078: Should there be an assert for the header value OR is it fine for the header value to be empty?

source/sigv4.c
```diff
@@ -794,6 +918,308 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+            keyLen = canonicalRequest->pHeadersLoc[ headerIndex ].key.dataLen;
+            valLen = canonicalRequest->pHeadersLoc[ headerIndex ].value.dataLen;
+            headerKey = canonicalRequest->pHeadersLoc[ headerIndex ].key.pData;
+            sigV4Status = copyHeaderStringToCanonicalBuffer( headerKey, keyLen, flags, ':', canonicalRequest );
```
aggarw13, line 1082: Please add a comment to mention that `:` is used to separate the header key and header value in the canonicalized form.

source/sigv4.c
```diff
@@ -794,6 +918,308 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+                canonicalRequest->pHeadersLoc[ noOfHeaders ].value.dataLen = ( pCurrLoc - pKeyOrValStartLoc );
+                /* Set starting location of the next header key string after the "\n". */
+                pKeyOrValStartLoc = pCurrLoc + 1U;
+                keyFlag = true;
+                noOfHeaders++;
+            }
+
+            pCurrLoc++;
+        }
+
```
aggarw13, line 1163: It would be important to verify that each header key has found its header value by checking that the `keyFlag` value is reset to `true` (unless it is fine to have headers field without any value).

source/sigv4.c
```diff
@@ -794,6 +918,308 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
...
+/*-----------------------------------------------------------*/
+
+    static SigV4Status_t appendAllHeadersToCanonicalRequest( const char * pHeaders,
+                                                             size_t headersLen,
+                                                             uint32_t flags,
+                                                             CanonicalContext_t * canonicalRequest )
+    {
+        bool keyFlag = true;
+        size_t noOfHeaders = 0;
```
aggarw13, line 1177:
```suggestion
        size_t noOfHeaders = 0;
```

source/sigv4.c
yourslab: According to here:
https://stackoverflow.com/questions/12130910/how-to-interpret-empty-http-accept-header

Looks like it's possible for the value to empty. Although it'd be pointless to have an empty header value if its value is meant to be signed for the purposes of SigV4.

source/sigv4.c
gshvang: When we are adding the value the separator will be '\n' and we don't want to lowercase the value.

source/sigv4.c
gshvang: I checked the coreHTTP library so when we add HTTP header, coreHTTP library does allows empty keys or values.

aggarw13: In that case, let's add a comment in the API documentation stating that empty header values (or header values containing whitespaces) are not supported

source/include/sigv4.h
```diff
@@ -44,6 +44,7 @@
  * defined in sigv4_config.h file. */
 #include "sigv4_config_defaults.h"
 
+#include "sigv4_internal.h"
```
aggarw13, line 47: Why is inclusion of this required?

source/include/sigv4.h
```diff
@@ -154,7 +155,10 @@ typedef enum SigV4Status
 
     /**
      * @brief The maximum number of header parameters was exceeded while parsing
-     * the http header string input parameter.
+     * the http header string input parameter. Default Maximum headers specified in the
+     * library is 100 as defined by the SIGV4_MAX_HTTP_HEADER_COUNT macro. It can be
+     * changed by defining SIGV4_MAX_HTTP_HEADER_COUNT macro in the custom config file
+     * provided by the application.
```
aggarw13, line 161:
```suggestion
     * the http header string passed to the library. 
     * The maximum number of supported HTTP headers can be configured
     * with the SIGV4_MAX_HTTP_HEADER_COUNT macro in the library config file
     * passed by the application. 
```

source/include/sigv4.h
```diff
@@ -367,6 +371,9 @@ typedef struct SigV4Parameters
 /**
  * @brief Generates the HTTP Authorization header value.
  *
+ * Note: Empty HTTP header keys or values are not supported which are passed
+ * as a part of SigV4Parameters_t.
```
aggarw13, line 375:
```suggestion
 * @note The API does not support HTTP headers containing empty HTTP header keys or values.
```

source/include/sigv4.h
gshvang: It was committed mistakenly. I included it for my own testing purposes.

aggarw13 (approved):

xortothemoon (dismissed):

yourslab (approved):

paulbartell (dismissed):

yourslab (approved):

source/sigv4.c
```diff
@@ -883,38 +883,6 @@ static SigV4Status_t getCredentialScope( SigV4Parameters_t * pSigV4Params,
         *canonicalURILen = index;
     }
 
-/*-----------------------------------------------------------*/
-
-    static void generateCanonicalURI( const char * pURI,
```
aggarw13, line 765: Why have the `verifySigV4Parameters`, `generateCanonicalURI` and `generateCanonicalQuery` functions been removed?

source/include/sigv4_internal.h
```diff
@@ -26,84 +26,384 @@
...
+        * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+        * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+        */
+
+/**
+ * @file sigv4_internal.h
+ * @brief Internal definitions for the SigV4 Client Utility Library.
+ */
+
+#ifndef SIGV4_INTERNAL_H_
```
aggarw13, line 232: There are two copies of the same header content present in the file

aggarw13 (approved):

paulbartell (approved):

PR #29, opened 2021-08-04
SigV4_GenerateHTTPAuthorization unit tests

yourslab:
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


test/unit-test/sigv4_utest.c
```diff
@@ -214,3 +493,209 @@ void test_SigV4_AwsIotDateToIso8601_Formatting_Error()
...
+{
+    SigV4Status_t returnStatus;
+    size_t i;
+
+    for( i = 0U; i < HAPPY_PATH_HASH_ITERATIONS; i++ )
+    {
+        resetFailableHashParams();
+        initHashCallToFail = i;
+        returnStatus = SigV4_GenerateHTTPAuthorization( &params, authBuf, &authBufLen, &signature, &signatureLen );
+        TEST_ASSERT_EQUAL( SigV4HashError, returnStatus );
```
muneebahmed10, line 674: Would it be useful to know which iteration failed? Perhaps by using `TEST_ASSERT_EQUAL_MESSAGE` to log the failing iteration?

test/unit-test/sigv4_utest.c
```diff
@@ -214,3 +493,209 @@ void test_SigV4_AwsIotDateToIso8601_Formatting_Error()
...
+void test_SigV4_GenerateHTTPAuthorization_Greater_Than_Max_Header_Query_Count()
+{
+    SigV4Status_t returnStatus;
+
+    params.pHttpParameters->pQuery = QUERY_STRING_GT_MAX_PARAMS;
+    params.pHttpParameters->queryLen = STR_LIT_LEN( QUERY_STRING_GT_MAX_PARAMS );
+    returnStatus = SigV4_GenerateHTTPAuthorization( &params, authBuf, &authBufLen, &signature, &signatureLen );
+    TEST_ASSERT_EQUAL( SigV4MaxQueryPairCountExceeded, returnStatus );
+
+    resetInputParams();
```
muneebahmed10, line 700: Is this reset needed here for another test that is meant to be written?

lexicon.txt
```diff
@@ -25,35 +26,42 @@ config
 const
 constness
 copydoc
+crypto
 currentparameter
+cyrpto
```
muneebahmed10, line 31:
```suggestion
```

test/unit-test/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 # Include filepaths for source and include.
 include( ${MODULE_ROOT_DIR}/sigv4FilePaths.cmake )
 
+find_package(OpenSSL REQUIRED)
+set(CMAKE_BUILD_TYPE "Debug")
+set(CMAKE_C_FLAGS "-Wall -Wextra -O0 -ggdb")
+
+add_definitions(-DNDEBUG)
```
yourslab, line 8: Probably only OpenSSL should be REQUIRED here. The rest of the flags I added just for testing things locally.

source/sigv4.c
```diff
@@ -1199,56 +1194,63 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
         assert( *canonicalURILen > 0U );
 
         pUriLoc = pUri;
-        pBufLoc = pCanonicalURI;
         bufferLen = *canonicalURILen;
 
-        while( ( i++ < uriLen ) && ( returnStatus == SigV4Success ) )
+        while( ( uriIndex++ < uriLen ) && ( bytesConsumed < bufferLen ) && ( returnStatus == SigV4Success ) )
```
muneebahmed10, line 1199: I got around this in my MISRA fixes by instead changing the `else if` code from
```c
*pBufLoc = *pUriLoc;
++pBufLoc;
++bytesConsumed;
```
to
```c
if( bytesConsumed < bufferLen )
{
    *pBufLoc = *pUriLoc;
    ++pBufLoc;
    ++bytesConsumed;
}
else
{
    returnStatus = SigV4InsufficientMemory;
    LOG_INSUFFICIENT_MEMORY_ERROR( "encode the URI", bytesConsumed + 1U - bufferLen );
}
```
This appropriately sets the returnStatus to `SigV4InsuffcientMemory` in the loop, instead of adding code later to deal specifically with that case

test/unit-test/sigv4_utest.c
```diff
@@ -21,19 +21,148 @@
...
+
+#define STR_LIT_LEN( LIT )    ( sizeof( LIT ) - 1U )
 
 /* The number of invalid date inputs tested in
  * test_SigV4_AwsIotDateToIso8601_Formatting_Error() */
-#define SIGV4_TEST_INVALID_DATE_COUNT    24U
+#define SIGV4_TEST_INVALID_DATE_COUNT                         24U
+
+#define AUTH_BUF_LENGTH                                       1000
+/*#define PATH                                       "/hi | world" */
```
muneebahmed10, line 40: Should this line be deleted?
```suggestion
```

test/unit-test/sigv4_utest.c
```diff
@@ -64,11 +193,191 @@ void formatAndVerifyInputDate( const char * pInputDate,
...
+    if( SHA256_Init( ( SHA256_CTX * ) pHashContext ) == 1 )
+    {
+        return 0;
+    }
+
+    return -1;
+}
+
+static int32_t valid_sha256_update( void * pHashContext,
+                                    const char * pInput,
```
muneebahmed10, line 209: Heads up, I'm changing the type of this in the crypto interface from `const char *` to `const uint8_t *` in my PR, so you  would have to change this here too if that goes in first.

test/unit-test/sigv4_utest.c
```diff
@@ -64,11 +193,191 @@ void formatAndVerifyInputDate( const char * pInputDate,
...
+    if( SHA256_Update( ( SHA256_CTX * ) pHashContext, pInput, inputLen ) )
+    {
+        return 0;
+    }
+
+    return -1;
+}
+
+static int32_t valid_sha256_final( void * pHashContext,
+                                   char * pOutput,
```
muneebahmed10, line 221: Here too

test/unit-test/sigv4_utest.c
```diff
@@ -64,11 +193,191 @@ void formatAndVerifyInputDate( const char * pInputDate,
...
+    {
+        return 0;
+    }
+
+    return -1;
+}
+
+/*==================== Echo Implementation of Crypto Interface ===================== */
+
+static hashEchoBuffer[ SIGV4_HASH_MAX_BLOCK_LENGTH ];
```
muneebahmed10, line 234: No type? It doesn't look like this is meant to be an `int`, which is the default
```suggestion
static uint8_t hashEchoBuffer[ SIGV4_HASH_MAX_BLOCK_LENGTH ];
```

test/unit-test/sigv4_utest.c
```diff
@@ -214,3 +523,468 @@ void test_SigV4_AwsIotDateToIso8601_Formatting_Error()
...
+    /* Test when the query string contains query parameters with exactly matching parameter names as well
+     * as matching values for those parameters. The query values of matching parameters differ in length though. */
+    params.pHttpParameters->pQuery = QUERY_MATCHING_PARAMS_AND_VALUES;
+    params.pHttpParameters->queryLen = strlen( QUERY_MATCHING_PARAMS_AND_VALUES );
+    returnStatus = SigV4_GenerateHTTPAuthorization( &params, authBuf, &authBufLen, &signature, &signatureLen );
+    TEST_ASSERT_EQUAL( SigV4Success, returnStatus );
+}
+
+/* Test that the library can handle HTTP headers data containing trimmable spaces. */
+void test_SigV4_GenerateAuthorization_Headers_With_Trimmable_Spaces()
```
muneebahmed10, line 727: Is there any way to verify the headers were trimmed properly?

test/unit-test/sigv4_utest.c
```diff
@@ -214,3 +523,468 @@ void test_SigV4_AwsIotDateToIso8601_Formatting_Error()
...
+    params.pHttpParameters->queryLen = STR_LIT_LEN( PRECANON_QUERY_TOO_LONG );
+    params.pHttpParameters->flags = SIGV4_HTTP_QUERY_IS_CANONICAL_FLAG;
+    returnStatus = SigV4_GenerateHTTPAuthorization( &params, authBuf, &authBufLen, &signature, &signatureLen );
+    TEST_ASSERT_EQUAL( SigV4InsufficientMemory, returnStatus );
+    /* END: Coverage for writeLineToCanonicalRequest(). */
+
+    /* Test case of insufficient memory when "String to Sign" cannot be stored in processing buffer.
+     * This scenario is produced by using a long AWS Region string (which is one of the parameters of String To Sign). */
+    char longRegion[ SIGV4_PROCESSING_BUFFER_LENGTH ];
+    /* Fill gibbering string data in the buffer for region. */
```
muneebahmed10, line 855: gibberish?

muneebahmed10 (changes requested):

test/unit-test/sigv4_utest.c
aggarw13: That's a good question. The only way in the current unit testing structure of only testing the public APIs is to test the signature generated. I am thinking of creating a separate PR for verifying the generated signatures for tests that expect successful generation of signature.

test/unit-test/CMakeLists.txt
aggarw13: Removed the flags

source/sigv4.c
aggarw13: Updated to use the nested if-else approach as complexity isn't a problem anymore

source/sigv4.c
```diff
@@ -1897,36 +1918,40 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
...
                 }
             }
 
-            if( ( remainingLen < 1U ) && ( numberOfParameters != ( i + 1U ) ) )
+            /* Check if there is a next parameter. */
+            thereExistsNextParameter = ( ( paramsIndex + 1U ) != numberOfParameters );
+
+            /* Before adding the '&' for the next query parameter, check that this is not
+             * the last parameter. */
+            if( thereExistsNextParameter && ( remainingLen > 0U ) )
```
muneebahmed10, line 1940: I would suggest this as it avoids the empty `else`
```c
if( thereExistsNextParameter )
{
    if( remainingLen > 0U )
    {
        *pBufLoc = '&';
        ++pBufLoc;
        remainingLen--;
    }
    else
    {
        returnStatus = SigV4InsufficientMemory;
        LOG_INSUFFICIENT_MEMORY_ERROR( "write the canonical query", 1U );
        break;
    }
}
```

source/sigv4.c
```diff
@@ -1243,57 +1237,59 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
         assert( canonicalURILen != NULL );
         assert( *canonicalURILen > 0U );
 
-        pUriLoc = pUri;
-        pBufLoc = pCanonicalURI;
         bufferLen = *canonicalURILen;
 
-        while( ( i++ < uriLen ) && ( returnStatus == SigV4Success ) )
+        for( ; ( uriIndex < uriLen ) && ( returnStatus == SigV4Success ); uriIndex++ )
```
muneebahmed10, line 1242: This is a MISRA 14.2 violations as a `for` loop since `returnStatus` is altered in the loop
```suggestion
        while( ( uriIndex++ < uriLen ) && ( returnStatus == SigV4Success ) )
```

source/sigv4.c
```diff
@@ -1243,57 +1237,59 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
...
-                    *pBufLoc = *pUriLoc;
-                    ++pBufLoc;
+                    pCanonicalURI[ bytesConsumed ] = pUri[ uriIndex ];
                     ++bytesConsumed;
                 }
                 else
                 {
                     returnStatus = SigV4InsufficientMemory;
-                    LOG_INSUFFICIENT_MEMORY_ERROR( "encode the URI", bytesConsumed + 1U - bufferLen );
+                    LogError( ( "Failed to encode URI in buffer due to insufficient memory" ) );
```
muneebahmed10, line 1269: Why remove the insufficient memory error macro? It's there to log this exact situation

muneebahmed10 (changes requested):

source/include/sigv4_internal.h
```diff
@@ -83,39 +83,42 @@
...
-#define LINEFEED_CHAR_LEN                      1U                                               /**< The length of #LINEFEED_CHAR. */
+#define LINEFEED_CHAR                          '\n'                                               /**< A linefeed character used to build the canonical request. */
+#define LINEFEED_CHAR_LEN                      1U                                                 /**< The length of #LINEFEED_CHAR. */
 
-#define SPACE_CHAR                             ' '                                              /**< A linefeed character used to build the Authorization header value. */
-#define SPACE_CHAR_LEN                         1U                                               /**< The length of #SPACE_CHAR. */
+#define CARRIAGE_RETURN_LINEFEED_STRING        "\r\n"                                             /**< The string used in non-canonicalized HTTP headers to separate header entries in HTTP request. */
+#define CARRIAGE_RETURN_LINEFEED_STRING_LEN    ( sizeof( CARRIAGE_RETURN_LINEFEED_STRING ) - 1U ) /**< The length of #CARRIAGE_RETURN_LINEFEED_STRING. */
```
muneebahmed10, line 95: The first 31 characters of these two macros are identical, which is a MISRA 5.4 violation

lexicon.txt
```diff
@@ -26,40 +27,52 @@ configpagestyle
 const
 constness
 copydoc
+<<<<<<< HEAD
```
abhidixi11, line 30: This looks wrong.

lexicon.txt
```diff
@@ -102,15 +117,24 @@ mon
...
+>>>>>>> origin/main
 opad
+openssl
 ored
 org
 outputlen
 paccesskeyid
 palgorithm
 param
+<<<<<<< HEAD
```
abhidixi11, line 133: Looks like merge issue.

lexicon.txt
```diff
@@ -102,15 +117,24 @@ mon
 monthsperday
 noninfringement
 nullterminate
+<<<<<<< HEAD
```
abhidixi11, line 120: Same - merge issue.

lexicon.txt
```diff
@@ -26,40 +27,52 @@ configpagestyle
...
 encodeslash
 encodetwice
 endif
 enums
 expirationlen
+failable
 feb
 formatchar
 formatlen
+<<<<<<< HEAD
```
abhidixi11, line 55: Merge issue.

source/sigv4.c
```diff
@@ -364,25 +365,29 @@ static void setQueryParameterValue( size_t currentParameter,
  * @param[in] keyLen The length of @p pKey.
  * @return Zero on success, all other return values are failures.
  */
-static int32_t hmacKey( HmacContext_t * pHmacContext,
-                        const char * pKey,
-                        size_t keyLen );
+static int32_t addKeyToHMAC( HmacContext_t * pHmacContext,
```
muneebahmed10, line 368: Should this keep the `hmac` so that all three hmac functions of `completeHmac` will have the same prefix?
```suggestion
static int32_t hmacAddKey( HmacContext_t * pHmacContext,
```

source/sigv4.c
```diff
@@ -1702,14 +1713,24 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
...
 
-        if( ( sigV4Status == SigV4Success ) && !FLAG_IS_SET( flags, SIGV4_HTTP_HEADERS_ARE_CANONICAL_FLAG ) )
+        if( sigV4Status == SigV4Success )
         {
-            /* Sorting headers based on keys. */
-            quickSort( canonicalRequest->pHeadersLoc, noOfHeaders, sizeof( SigV4KeyValuePair_t ), cmpHeaderField );
+            if( FLAG_IS_SET( flags, SIGV4_HTTP_HEADERS_ARE_CANONICAL_FLAG ) )
+            {
+                /* Headers are already canonicalized, so just write it to the buffer as is. */
+                sigV4Status = writeLineToCanonicalRequest( pHeaders,
```
muneebahmed10, line 1721: I thought if the headers are already canonicalized then we don't need to write them to the buffer right here? @gshvang, is that still the case?

source/sigv4.c
```diff
@@ -1812,31 +1833,39 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
...
 
-                    /* The previous field did not have a value set for it, so set its value to NULL. */
+                    /* Store the previous parameter's empty value information. Use NULL to represent empty value. */
                     setQueryParameterValue( currentParameter, NULL, 0U, pCanonicalRequest );
 
                     startOfFieldOrValue = i + 1U;
                     currentParameter++;
+
+                    /* Reset the parameter value state of the new parameter entry. */
+                    fieldHasValue = false;
```
muneebahmed10, line 1849: This looks like a no op to me. Why have it?

source/sigv4.c
```diff
@@ -1867,24 +1896,20 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
...
-            if( pValue != NULL )
-            {
-                returnStatus = encodeURI( pValue,
-                                          valueLen,
-                                          pBufCur + 1U,
-                                          &bytesWritten,
-                                          true,
-                                          true );
-            }
+            returnStatus = encodeURI( pValue,
```
muneebahmed10, line 1907: Does the comment about the value being empty no longer apply?

source/sigv4.c
aggarw13: Sure

source/include/sigv4_internal.h
aggarw13: Renamed the macros

source/sigv4.c
aggarw13: Adding nesting increases complexity

source/sigv4.c
aggarw13: Removed

source/sigv4.c
```diff
@@ -2812,15 +2867,25 @@ static SigV4Status_t generateSigningKey( const SigV4Parameters_t * pSigV4Params,
...
+        returnStatus = SigV4InsufficientMemory;
+    }
+    else if( hmacStatus != 0 )
     {
         returnStatus = SigV4HashError;
     }
+    else
+    {
+        /* Empty else for MISRA C:2012 compliance. */
+    }
```
muneebahmed10, line 2888: Since `isBufferSpaceSufficient` is only ever set once, you could just set the return status at the same time. Then this block could just be
```suggestion
    /* Set the error code for hash failures. */
    if( hmacStatus != 0 )
    {
        returnStatus = SigV4HashError;
    }
```
and you wouldn’t need the empty else

source/sigv4.c
```diff
@@ -2913,38 +2978,37 @@ SigV4Status_t SigV4_GenerateHTTPAuthorization( const SigV4Parameters_t * pParams
...
+        {
+            /* The default algorithm is AWS4-HMAC-SHA256. */
+            pAlgorithm = SIGV4_AWS4_HMAC_SHA256;
+            algorithmLen = SIGV4_AWS4_HMAC_SHA256_LENGTH;
+        }
+        else
+        {
+            pAlgorithm = pParams->pAlgorithm;
+            algorithmLen = pParams->algorithmLen;
+        }
```
muneebahmed10, line 3004: If the complexity of this function is too high, you can reduce it by moving this code out of the `if( returnStatus == SigV4Success )` block

source/sigv4_quicksort.c
```diff
@@ -144,9 +144,30 @@ static void quickSortHelper( void * pArray,
...
-        len2 = ( ( partitionIndex + 1U ) < hi ) ? ( hi - partitionIndex - 1U ) : 0U;
-
+        /* Calculate length of the left partition containing items smaller
+         * than the pivot element.
+         * The length is zero if either:
+         * 1. The pivoted item is the smallest in the
+         * the array before partitioning.
+         *              OR
+         * 2. The left partition is only of single length which can be treated as
+         * sorted, and thus, of zero length for avoided adding to the stack. */
```
muneebahmed10, line 154: You could alternatively describe this as
```
Calculate the difference between the first and last indices of the left partition.
Partitions with more than one element are not guaranteed to be sorted so must be
pushed onto the stack. If the pivot is either the smallest or second smallest item
in the array, the left partition is sorted.
```

source/sigv4_quicksort.c
```diff
@@ -144,9 +144,30 @@ static void quickSortHelper( void * pArray,
...
+        len1 = ( ( partitionIndex != 0U ) && ( ( partitionIndex - 1U ) > lo ) ) ? ( partitionIndex - lo ) : 0U;
+
+        /* Calculate length of the right partition containing items greater than
+         * or equal to the pivot item.
+         * The calculated length is zero if either:
+         * 1. The pivoted item is the smallest in the the array before partitioning.
```
muneebahmed10, line 160: greatest

muneebahmed10 (changes requested):

source/sigv4.c
muneebahmed10: Looks like this line was just moved here from elsewhere

source/sigv4.c
```diff
@@ -2741,27 +2787,36 @@ static SigV4Status_t generateSigningKey( const SigV4Parameters_t * pSigV4Params,
...
+    if( isBufferSpaceSufficient )
+    {
+        /* Fill the key prefix, "AWS4" in the HMAC context cache.
+         * The prefix is part of the key for the first round of HMAC operation:
+         * HMAC("AWS4" + SecretKey, Date) */
+        hmacStatus = hmacAddKey( pHmacContext,
+                                 SIGV4_HMAC_SIGNING_KEY_PREFIX,
+                                 SIGV4_HMAC_SIGNING_KEY_PREFIX_LEN,
+                                 true /* Is key prefix. */ );
+        assert( hmacStatus == 0 );
```
yourslab, line 2816: Why assert this? That may certainly not be the case depending on how the hash functions are implemented.

source/sigv4.c
```diff
@@ -2741,27 +2787,36 @@ static SigV4Status_t generateSigningKey( const SigV4Parameters_t * pSigV4Params,
     SigV4Status_t returnStatus = SigV4Success;
     int32_t hmacStatus = 0;
     char * pSigningKeyStart = NULL;
+    bool isBufferSpaceSufficient = true;
```
yourslab, line 2790: Is an extra flag really necessary here? One can just check the current value of `returnStatus`.

source/sigv4.c
yourslab: Because I did this already:
```
/**
 * @brief Verify all authorization-related parameters passed to
 * #SigV4_GenerateHTTPAuthorization.
 *
 * @param[in] pParams Complete SigV4 configurations passed by application.
 * @param[out] pAlgorithm The algorithm used for SigV4 authentication.
 * @param[out] algorithmLen The length of @p pAlgorithm.
 */
static void assignDefaultArguments( const SigV4Parameters_t * pParams,
                                    char ** pAlgorithm,
                                    size_t * algorithmLen );
```

```
static void assignDefaultArguments( const SigV4Parameters_t * pParams,
                                    char ** pAlgorithm,
                                    size_t * algorithmLen )
{
    if( ( pParams->pAlgorithm == NULL ) || ( pParams->algorithmLen == 0 ) )
    {
        /* The default algorithm is AWS4-HMAC-SHA256. */
        *pAlgorithm = SIGV4_AWS4_HMAC_SHA256;
        *algorithmLen = SIGV4_AWS4_HMAC_SHA256_LENGTH;
    }
    else
    {
        *pAlgorithm = pParams->pAlgorithm;
        *algorithmLen = pParams->algorithmLen;
    }
}
```

test/CMakeLists.txt
```diff
@@ -21,73 +18,69 @@ get_filename_component( __MODULE_ROOT_DIR "${CMAKE_CURRENT_LIST_DIR}/.." ABSOLUT
...
-    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-)
+    # Add a target for running coverage on tests.
+    add_custom_target( coverage
+        COMMAND ${CMAKE_COMMAND} -DCMOCK_DIR=${CMOCK_DIR}
+        -P ${MODULE_ROOT_DIR}/tools/cmock/coverage.cmake
+        DEPENDS cmock unity sigv4_utest
+        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+    )
+endif()
```
yourslab, line 86: Missing newline

source/sigv4.c
```diff
@@ -1929,36 +1970,43 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
...
+                remainingLen--;
             }
 
-            if( returnStatus != SigV4Success )
+            /* There is at least another query parameter entry to encode
+             * but no space in the buffer. */
+            else if( thereExistsNextParameter )
             {
+                returnStatus = SigV4InsufficientMemory;
+                LogError( ( "Unable to write canonical query: Insufficient memory configured in \"SIGV4_PROCESSING_BUFFER_LENGTH\"" ) );
```
yourslab, line 2004: Why enclose this in quotation marks? Better to refer to it as a macro and to mention about how much memory is lacking.

source/sigv4.c
```diff
@@ -351,45 +351,61 @@ static void setQueryParameterValue( size_t currentParameter,
...
  * @param[in] dataLen The length of @p pData.
+ *
  * @return Zero on success, all other return values are failures.
  */
-static int32_t hmacData( HmacContext_t * pHmacContext,
-                         const char * pData,
-                         size_t dataLen );
+static int32_t hmacIntermediate( HmacContext_t * pHmacContext,
+                                 const char * pData,
+                                 size_t dataLen );
```
yourslab, line 395: The whole point of this naming scheme was to match the naming schemes of the hash interface.

yourslab: Also, the original terminology is more commonly used in cryptographic operations

test/CMakeLists.txt
```diff
@@ -21,73 +18,69 @@ get_filename_component( __MODULE_ROOT_DIR "${CMAKE_CURRENT_LIST_DIR}/.." ABSOLUT
...
 
 #  ====================== Coverity Analysis Configuration ======================
-
 # Include filepaths for source and include.
 include( ${MODULE_ROOT_DIR}/sigv4FilePaths.cmake )
 
 # Target for Coverity analysis that builds the library.
 add_library( coverity_analysis
-             ${SIGV4_SOURCES} )
+${SIGV4_SOURCES} )
```
yourslab, line 36: Spacing weird here

test/unit-test/sigv4_utest.c
```diff
@@ -214,3 +532,725 @@ void test_SigV4_AwsIotDateToIso8601_Formatting_Error()
...
+ */
+void test_SigV4_GenerateHTTPAuthorization_Hash_Errors()
+{
+    SigV4Status_t returnStatus;
+    size_t i;
+
+    params.pCredentials->pSecretAccessKey = SECRET_KEY_LONGER_THAN_HASH_BLOCK;
+    params.pCredentials->secretAccessKeyLen = strlen( SECRET_KEY_LONGER_THAN_HASH_BLOCK );
+
+    for( i = 0U; i < HAPPY_PATH_HASH_ITERATIONS; i++ )
```
yourslab, line 1190: This is certainly a hack for now. I'd just leave a comment if you run out of time in actually being able to assert what iteration you expect to fail.

source/sigv4.c
```diff
@@ -2338,12 +2393,12 @@ static int32_t hmacFinal( HmacContext_t * pHmacContext,
...
-         * the inner-padded key then XOR with opad. XOR is associative,
-         * so one way to do this is by performing XOR on each byte of the
-         * inner-padded key with (0x36 ^ 0x5c) = (ipad ^ opad) = 0x6a.  */
+         * the inner-padded key (by XORing with ipad byte) and then XOR with opad
+         * to generate the outer-padded key. As XOR is associative, one way to do
+         * this is by performing XOR on each byte of the inner-padded key (ipad ^ opad).  */
         for( i = 0U; i < pCryptoInterface->hashBlockLen; i++ )
         {
-            pHmacContext->key[ i ] ^= 0x6aU;
+            pHmacContext->key[ i ] ^= ( HMAC_INNER_PAD_BYTE ^ HMAC_OUTER_PAD_BYTE );
```
yourslab, line 2401: This may XOR ipad and opad constants on each iteration depending on the compiler. Better to have a macro for `0x6aU`.

source/sigv4.c
```diff
@@ -351,45 +351,61 @@ static void setQueryParameterValue( size_t currentParameter,
...
+ *
  * @return Zero on success, all other return values are failures.
  */
-static int32_t hmacKey( HmacContext_t * pHmacContext,
-                        const char * pKey,
-                        size_t keyLen );
+static int32_t hmacAddKey( HmacContext_t * pHmacContext,
+                           const char * pKey,
+                           size_t keyLen,
+                           bool isKeyPrefix );
```
yourslab, line 376: Why bother with another input parameter?

source/sigv4.c
```diff
@@ -599,13 +612,22 @@ static SigV4Status_t parseDate( const char * pDate,
...
 /**
- * @brief Verify @p pParams and its sub-members.
+ * @brief Verify input parameters to the SigV4_GenerateHTTPAuthorization API.
  *
  * @param[in] pParams Complete SigV4 configurations passed by application.
+ * @param[in] pAuthBuf The user-supplied buffer for filling Authorization Header.
+ * @param[in] authBufLen The user-supplied size value of @p pAuthBuf buffer.
+ * @param[in] pSignature The user-supplied pointer memory to store starting location of
+ * Signature in Authorization Buffer.
+ * @param[in] signatureLen The user supplied pointer to store length of Signature.
```
yourslab, line 622: For consistency
```suggestion
 * @param[in] signatureLen The user-supplied pointer to store length of Signature.
```

test/unit-test/sigv4_utest.c
```diff
@@ -64,11 +209,184 @@ void formatAndVerifyInputDate( const char * pInputDate,
...
+    hashInputLen = inputLen;
+    ( void ) memcpy( hashEchoBuffer, pInput, inputLen );
+}
+
+static int32_t echo_hash_final( void * pHashContext,
+                                uint8_t * pOutput,
+                                size_t outputLen )
+{
+    ( void ) memcpy( pOutput, hashEchoBuffer, hashInputLen );
+}
```
yourslab, line 274: I don't think the echo implementation is as useful as I initially thought and should probably just be removed

test/unit-test/sigv4_utest.c
```diff
@@ -64,11 +209,184 @@ void formatAndVerifyInputDate( const char * pInputDate,
...
+
+/*==================== Failable Implementation of Crypto Interface ===================== */
+
+#define HAPPY_PATH_HASH_ITERATIONS    12U
```
yourslab, line 278: This doesn't sound like an appropriate name to use?

test/unit-test/sigv4_utest.c
```diff
@@ -214,3 +532,725 @@ void test_SigV4_AwsIotDateToIso8601_Formatting_Error()
...
+    params.pHttpParameters->pPath = NULL_TERMINATED_PATH;
+    params.pHttpParameters->pathLen = NULL_TERMINATED_PATH_LEN;
+    returnStatus = SigV4_GenerateHTTPAuthorization( &params, authBuf, &authBufLen, &signature, &signatureLen );
+    TEST_ASSERT_EQUAL( SigV4Success, returnStatus );
+
+    /* Coverage for double-encoded equals in query string value. */
+    params.pHttpParameters->pQuery = QUERY_VALUE_HAS_EQUALS;
+    params.pHttpParameters->queryLen = STR_LIT_LEN( QUERY_VALUE_HAS_EQUALS );
+    returnStatus = SigV4_GenerateHTTPAuthorization( &params, authBuf, &authBufLen, &signature, &signatureLen );
+    TEST_ASSERT_EQUAL( SigV4Success, returnStatus );
```
yourslab, line 704: So we never assert the actual signature matches what we expect. That's quite useful to have for a test, but I guess we've run out of time at this point.

yourslab: Or it would be good to have at least one test case that asserts the signature. Otherwise, this is really just a test of return codes for the sake of coverage...

source/sigv4.c
aggarw13: Adding the prefix does not involve hashing operations but only involves adding the prefix string to the cache. Thus, I added the assert.

test/unit-test/sigv4_utest.c
aggarw13: Removed

source/sigv4.c
aggarw13: Because we want to add the zero padding only when the key is completely received. I thought it is better to encapsulate the `K'` operation within this function than distribute it across two functions

source/sigv4.c
aggarw13: I felt that using terminology that follows the HMAC algorithmic operations is easier for a developer to make sense of the relationship between the `hmac*` internal functions. As these are internal functions, I prefer prioritize ease of readability

test/unit-test/sigv4_utest.c
aggarw13: Added

source/include/sigv4_internal.h
```diff
@@ -28,6 +28,8 @@
 #ifndef SIGV4_INTERNAL_H_
 #define SIGV4_INTERNAL_H_
 
+#include <stdbool.h>
```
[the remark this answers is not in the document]
aggarw13: Not anymore

abhidixi11 (approved):
I'm approving it based on the decision that unit test code quality related comments will be addressed in subsequent PR.

test/unit-test/sigv4_utest.c
aggarw13: Yes, updated the name

source/sigv4_quicksort.c
```diff
@@ -202,31 +201,32 @@ static size_t partition( void * pArray,
                          size_t itemSize,
                          ComparisonFunc_t comparator )
 {
-    void * pivot;
+    uint8_t * pivot;
+    uint8_t * pArrayLocal = ( uint8_t * ) pArray;
```
aggarw13, line 205: This addresses the warnings generated (by `gcc-7.5.0`) for using arithmetic operations on `void *` 
```
/home/ubuntu/Repos/SigV4-for-AWS-IoT-embedded-sdk/source/sigv4_quicksort.c:212:20: warning: pointer of type ‘void *’ used in arithmetic [-Wpointer-arith]
     pivot = pArray + ( high * itemSize );
                    ^
/home/ubuntu/Repos/SigV4-for-AWS-IoT-embedded-sdk/source/sigv4_quicksort.c:220:32: warning: pointer of type ‘void *’ used in arithmetic [-Wpointer-arith]
         if( comparator( pArray + ( j * itemSize ), pivot ) < 0 )
                                ^
/home/ubuntu/Repos/SigV4-for-AWS-IoT-embedded-sdk/source/sigv4_quicksort.c:223:26: warning: pointer of type ‘void *’ used in arithmetic [-Wpointer-arith]
             swap( pArray + ( i * itemSize ), pArray + ( j * itemSize ), itemSize );
                          ^
/home/ubuntu/Repos/SigV4-for-AWS-IoT-embedded-sdk/source/sigv4_quicksort.c:223:53: warning: pointer of type ‘void *’ used in arithmetic [-Wpointer-arith]
             swap( pArray + ( i * itemSize ), pArray + ( j * itemSize ), itemSize );
                                                     ^
/home/ubuntu/Repos/SigV4-for-AWS-IoT-embedded-sdk/source/sigv4_quicksort.c:229:18: warning: pointer of type ‘void *’ used in arithmetic [-Wpointer-arith]
     swap( pArray + ( ( i + 1U ) * itemSize ), pivot, itemSize );
                  ^
```

source/sigv4.c
yourslab: It would be good to rename the hash interface then to match this new naming scheme

yourslab: It would be good to rename the hash interface then to match this new naming scheme

muneebahmed10 (approved):

PR #32, opened 2021-08-11
Add some MISRA fixes

muneebahmed10:
*Description*:

Fixes about a third of the warnings from static analysis. The remaining ones are mostly due to signed/unsigned conversions and the use of ctype functions/qsort.
These commits fix most of the following warnings:
- Fix overrun error and MISRA 15.7, 17.8, 8.3, 8.13, 11.8
- Fix MISRA 2.2, 13.3, 14.2, and 14.3
- Fix MISRA 12.1

### DIFF
--- a/source/sigv4.c
+++ b/source/sigv4.c
@@ -884,7 +884,7 @@ static SigV4Status_t parseDate( const char * pDate,
             formatIndex++;
 
             /* Numerical value of length specifier character. */
-            lenToRead = ( ( uint32_t ) pFormat[ formatIndex ] - ( uint32_t ) '0' );
+            lenToRead = ( size_t ) ( pFormat[ formatIndex ] - '0' );
             formatIndex++;
 
             /* Ensure read is within buffer bounds. */
@@ -1206,8 +1206,8 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
         {
             if( doubleEncodeEquals && ( *pUriLoc == '=' ) )
             {
-                if( ( bytesConsumed > SIZE_MAX - URI_DOUBLE_ENCODED_EQUALS_CHAR_SIZE ) ||
-                    ( bytesConsumed + URI_DOUBLE_ENCODED_EQUALS_CHAR_SIZE > bufferLen ) )
+                if( ( bytesConsumed > ( SIZE_MAX - URI_DOUBLE_ENCODED_EQUALS_CHAR_SIZE ) ) ||
+                    ( ( bytesConsumed + URI_DOUBLE_ENCODED_EQUALS_CHAR_SIZE ) > bufferLen ) )
                 {
                     returnStatus = SigV4InsufficientMemory;
                     LOG_INSUFFICIENT_MEMORY_ERROR( "encode the URI",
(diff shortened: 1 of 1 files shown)


source/sigv4.c
```diff
@@ -2572,23 +2599,20 @@ static SigV4Status_t generateAuthorizationValuePrefix( const SigV4Parameters_t *
...
-            returnStatus = SigV4InsufficientMemory;
-        }
+    if( *pAuthPrefixLen < ( authPrefixLen + encodedSignatureLen ) )
+    {
+        LogError( ( "Insufficient memory provided to write the Authorization header value, bytesExceeded=%lu",
+                    ( unsigned long ) ( authPrefixLen + encodedSignatureLen - *pAuthPrefixLen ) ) );
+        returnStatus = SigV4InsufficientMemory;
     }
 
     /* START:  Writing of authorization value prefix. */
```
gshvang, line 2618: We should check return status here before writing the content to buffer.

source/sigv4.c
```diff
@@ -2455,16 +2485,13 @@ static SigV4Status_t generateCanonicalRequestUntilHeaders( const SigV4Parameters
...
-        returnStatus = writeLineToCanonicalRequest( pParams->pHttpParameters->pHttpMethod,
-                                                    pParams->pHttpParameters->httpMethodLen,
-                                                    pCanonicalContext );
-    }
+    /* Write the HTTP Request Method to the canonical request. */
+    returnStatus = writeLineToCanonicalRequest( pParams->pHttpParameters->pHttpMethod,
+                                                pParams->pHttpParameters->httpMethodLen,
+                                                pCanonicalContext );
 
     /* Set defaults for path and algorithm. */
```
gshvang, line 2496: Return status should be checked here also.

muneebahmed10: Return status is checked on line 2510

source/sigv4.c
```diff
@@ -69,26 +69,37 @@
...
+ * @param[in, out] pCanonicalContext Struct to maintain intermediary buffer
  * and state of canonicalization.
  */
     static SigV4Status_t generateCanonicalQuery( const char * pQuery,
                                                  size_t queryLen,
-                                                 CanonicalContext_t * canonicalRequest );
+                                                 CanonicalContext_t * pCanonicalContext );
+
+/**
+ * @brief Determine if a character can be written without needing encoding.
```
aggarw13, line 94:
```suggestion
 * @brief Determine if a character can be written without needing URI encoding when generating Canonical Request.
```

source/sigv4.c
gshvang: I think we should check here also for the lines 2470-2481. Because if the status returned line 2492 is not success we don't need to execute the lines 2470-2481.

muneebahmed10: The additional check would be more trouble than its worth since those lines only set two variables that were declared on the function's stack and don't affect the input. I can rearrange those lines to come before this one, but they will still get executed

gshvang: Its better to rearrange those lines.

gshvang (dismissed):

source/sigv4.c
```diff
@@ -1178,6 +1189,14 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
         return URI_DOUBLE_ENCODED_EQUALS_CHAR_SIZE;
     }
 
+/*-----------------------------------------------------------*/
+
+    static bool isAllowedChar( char c,
+                               bool encodeSlash )
+    {
+        return( isalnum( c ) || ( c == '-' ) || ( c == '_' ) || ( c == '.' ) || ( c == '~' ) || ( ( c == '/' ) && !encodeSlash ) );
```
aggarg, line 1197:
```suggestion
        return( isalnum( c ) || ( c == '-' ) || ( c == '_' ) || ( c == '.' ) || ( c == '~' ) || ( ( c == '/' ) && ( encodeSlash == false ) ) );
```

aggarg (approved):

yourslab (approved):

source/sigv4.c
muneebahmed10: Since I'm touching this line in #33 as well, I think I'll address it then. This line also uses `isalnum`, which we might have to reimplement to comply with MISRA directive 4.6, so this might be converted into an `if-else if-else` block anyway

PR #33, opened 2021-08-12
Add more MISRA fixes

muneebahmed10:
*Description*:
Addresses more warnings from static analysis, mostly due to unsigned/signed/boolean conversions. ~Ignore the first commit until #32 is merged as this PR is based on top of those changes~

- Update some char * to uint8_t *
- Fix 10.4 warnings
- Add macro for bit flags
- Add ptrdiff_t for 10.8 and 10.3 warnings
- Address 7.4 warnings
- Address some ctype warnigs

### DIFF
--- a/source/include/sigv4_internal.h
+++ b/source/include/sigv4_internal.h
@@ -119,6 +119,11 @@
     }
 
 /**
+ * @brief A helper macro to test if a flag is set.
+ */
+#define FLAG_IS_SET( bits, flag )    ( ( ( bits ) & ( flag ) ) == ( flag ) )
+
+/**
  * @brief An aggregator representing the individually parsed elements of the
  * user-provided date parameter. This is used to verify the complete date
  * representation, and construct the final ISO 8601 string.
--- a/source/include/sigv4_internal.h
+++ b/source/include/sigv4_internal.h
@@ -113,10 +113,10 @@
  * @brief A helper macro to print insufficient memory errors.
  */
 #define LOG_INSUFFICIENT_MEMORY_ERROR( purposeOfWrite, bytesExceeded )                       \
-    do {                                                                                     \
+    {                                                                                        \
         LogError( ( "Insufficient memory provided to " purposeOfWrite ", bytesExceeded=%lu", \
                     ( unsigned long ) ( bytesExceeded ) ) );                                 \
-    } while( 0 )
+    }
 
 /**
  * @brief An aggregator representing the individually parsed elements of the
@@ -196,7 +196,7 @@ typedef struct HmacContext
     /**
      * @brief All accumulated key data.
      */
-    char key[ SIGV4_HASH_MAX_BLOCK_LENGTH ];
+    uint8_t key[ SIGV4_HASH_MAX_BLOCK_LENGTH ];
 
     /**
      * @brief The length of the accumulated key data.
--- a/source/sigv4.c
+++ b/source/sigv4.c
@@ -362,7 +362,7 @@ static void setQueryParameterValue( size_t currentParameter,
  * @return Zero on success, all other return values are failures.
  */
 static int32_t hmacKey( HmacContext_t * pHmacContext,
-                        const char * pKey,
+                        const uint8_t * pKey,
                         size_t keyLen );
 
 /**
@@ -432,9 +432,9 @@ static int32_t completeHmac( HmacContext_t * pHmacContext,
  * @param[in] pCryptoInterface The interface used to call hash functions.
  * @return Zero on success, all other return values are failures.
  */
-static int32_t completeHash( const char * pInput,
+static int32_t completeHash( const uint8_t * pInput,
                              size_t inputLen,
-                             char * pOutput,
+                             uint8_t * pOutput,
                              size_t outputLen,
                              const SigV4CryptoInterface_t * pCryptoInterface );
 
@@ -1359,12 +1359,12 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
         if( isspace( value[ index ] ) )
         {
             /* The last character is a trailing space. */
-            if( ( index + 1 ) == valLen )
+            if( ( index + 1U ) == valLen )
             {
                 ret = true;
             }
             /* Trim if the next character is also a space. */
-            else if( isspace( value[ index + 1 ] ) )
+            else if( isspace( value[ index + 1U ] ) )
             {
                 ret = true;
             }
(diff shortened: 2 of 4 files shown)


gshvang (approved):

source/include/sigv4_internal.h
```diff
@@ -113,10 +113,15 @@
  * @brief A helper macro to print insufficient memory errors.
  */
 #define LOG_INSUFFICIENT_MEMORY_ERROR( purposeOfWrite, bytesExceeded )                       \
-    do {                                                                                     \
+    {                                                                                        \
         LogError( ( "Insufficient memory provided to " purposeOfWrite ", bytesExceeded=%lu", \
                     ( unsigned long ) ( bytesExceeded ) ) );                                 \
-    } while( 0 )
+    }
```
muneebahmed10, line 119: This was removed due to a MISRA warning that `0` is not a boolean, but if we want to keep the `do-while`, then it can be a constant condition:
```suggestion
    } while( 1 == 0 )
```

yourslab: Should we keep the do-while? It's a little unconventional that just using what we have now.

source/sigv4.c
```diff
@@ -625,6 +627,20 @@ static SigV4Status_t lowercaseHexEncode( const SigV4String_t * pInputStr,
...
 
+/**
+ * @brief Copy a string into a char * buffer.
+ * @note This function can be used to copy a string literal without
+ * MISRA warnings.
+ *
+ * @param[in] destination The buffer to write
+ * @param[in] source String to copy.
+ * @param[in] length Number of characters to copy.
+ * @return @p length
```
yourslab, line 638: I think `@return` needs to be elaborated a bit more

source/sigv4.c
```diff
@@ -966,9 +985,9 @@ static SigV4Status_t lowercaseHexEncode( const SigV4String_t * pInputStr,
     {
         for( i = 0; i < pInputStr->dataLen; i++ )
         {
-            *hex = digitArr[ ( pInputStr->pData[ i ] & 0xF0 ) >> 4 ];
+            *hex = digitArr[ ( bytes[ i ] & 0xF0U ) >> 4 ];
             hex++;
-            *hex = digitArr[ ( pInputStr->pData[ i ] & 0x0F ) ];
+            *hex = digitArr[ ( bytes[ i ] & 0x0FU ) ];
```
yourslab, line 990: Why do we need an additional variable here? Just preference?

source/sigv4.c
```diff
@@ -1356,15 +1377,15 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
         assert( ( value != NULL ) && ( index < valLen ) );
 
         /* Only trim spaces. */
-        if( isspace( value[ index ] ) )
+        if( isspace( value[ index ] ) != 0U )
```
yourslab, line 1380: This will have its own implementation in a separate PR

source/sigv4.c
muneebahmed10: It's since MISRA requires all the operands of bitwise operators to be unsigned, so I had to either add an unsigned variable or do `( ( ( uint8_t ) pInputStr->pData[ i ] ) & 0x0FU )`.

source/sigv4.c
muneebahmed10: Does it really? The third parameter is the return value, unchanged.

source/sigv4.c
muneebahmed10: True, I just wanted wanted one of the warnings to go away. It'll be overwritten later anyway

source/sigv4.c
```diff
@@ -1225,8 +1246,7 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
         {
             if( doubleEncodeEquals && ( *pUriLoc == '=' ) )
             {
-                if( ( bytesConsumed > ( SIZE_MAX - URI_DOUBLE_ENCODED_EQUALS_CHAR_SIZE ) ) ||
-                    ( ( bytesConsumed + URI_DOUBLE_ENCODED_EQUALS_CHAR_SIZE ) > bufferLen ) )
+                if( ( bufferLen - bytesConsumed ) < URI_DOUBLE_ENCODED_EQUALS_CHAR_SIZE )
```
yourslab, line 1249: What happens if `bytesConsumed` > `bufferLen`?

source/sigv4.c
```diff
@@ -1239,14 +1259,21 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
...
+                {
+                    returnStatus = SigV4InsufficientMemory;
+                    LOG_INSUFFICIENT_MEMORY_ERROR( "encode the URI", bytesConsumed + 1U - bufferLen );
+                }
             }
             else
             {
-                if( ( bytesConsumed > ( SIZE_MAX - URI_ENCODED_SPECIAL_CHAR_SIZE ) ) ||
-                    ( ( bytesConsumed + URI_ENCODED_SPECIAL_CHAR_SIZE ) > bufferLen ) )
+                if( ( bufferLen - bytesConsumed ) < URI_ENCODED_SPECIAL_CHAR_SIZE )
```
yourslab, line 1276: We should probably check `bufferLen` > `bytesConsumed`

muneebahmed10: No, since `bytesConsumed` is initialized to 0 and `bufferLen` is initialized to `*canonicalURILen` which is >= 0 as these are all `size_t`s. In every branch there is a check that there is enough space in the buffer before writing any bytes, so it's never possible for `bytesConsumed > bufferLen`

source/sigv4.c
muneebahmed10: See my other comment. The `else if` block was altered to make sure that this is not possible

source/sigv4.c
yourslab: Okay, I suppose CBMC will just catch if there are any misses

yourslab (approved):

source/sigv4.c
```diff
@@ -625,6 +627,20 @@ static SigV4Status_t lowercaseHexEncode( const SigV4String_t * pInputStr,
  */
 static size_t sizeNeededForCredentialScope( const SigV4Parameters_t * pSigV4Params );
 
+/**
+ * @brief Copy a string into a char * buffer.
+ * @note This function can be used to copy a string literal without
+ * MISRA warnings.
+ *
+ * @param[in] destination The buffer to write
```
aggarg, line 635:
```suggestion
 * @param[in] destination The buffer to write.
```

source/sigv4.c
aggarg: If the third parameter is always returned, what is the point of returning it then?

source/sigv4.c
```diff
@@ -1194,7 +1215,7 @@ static SigV4Status_t generateCredentialScope( const SigV4Parameters_t * pSigV4Pa
     static bool isAllowedChar( char c,
                                bool encodeSlash )
     {
-        return( isalnum( c ) || ( c == '-' ) || ( c == '_' ) || ( c == '.' ) || ( c == '~' ) || ( ( c == '/' ) && !encodeSlash ) );
+        return( ( isalnum( c ) != 0U ) || ( c == '-' ) || ( c == '_' ) || ( c == '.' ) || ( c == '~' ) || ( ( c == '/' ) && !encodeSlash ) );
```
aggarg, line 1218:
```suggestion
        return( ( isalnum( c ) != 0U ) || ( c == '-' ) || ( c == '_' ) || ( c == '.' ) || ( c == '~' ) || ( ( c == '/' ) && ( encodeSlash == false ) ) );
```

source/sigv4.c
muneebahmed10: Everywhere this function is used, the buffer is incremented by the number of bytes written. Making it the return value means I can write
```c
pBuffer += copyString( pBuffer, str, length );
```
instead of
```c
copyString( pBuffer, str, length );
pBuffer += length;
```
which reduces complexity a little

aggarg (approved):

yourslab (approved):

PR #45, opened 2021-08-16
Add C++ guards to SigV4 headers

yourslab:
Allows C++ linkage by adding guards to header files

### DIFF
--- a/source/include/sigv4_internal.h
+++ b/source/include/sigv4_internal.h
@@ -28,6 +28,12 @@
 #ifndef SIGV4_INTERNAL_H_
 #define SIGV4_INTERNAL_H_
 
+/* *INDENT-OFF* */
+#ifdef __cplusplus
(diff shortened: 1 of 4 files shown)


gshvang (approved):

source/include/sigv4_internal.h
```diff
@@ -28,6 +28,12 @@
 #ifndef SIGV4_INTERNAL_H_
 #define SIGV4_INTERNAL_H_
 
+/* *INDENT-OFF* */
+#ifdef __cplusplus
```
muneebahmed10, line 32: I don't think internal headers really need this

muneebahmed10 (approved):

source/include/sigv4_internal.h
yourslab: Yeah, probably not, doesn't hurt though

PR #51, opened 2021-08-17
[SigV4]Payload optimization

gshvang:
This PR contains changes related to optimizing the sigV4 library for the already hashed payload.

### DIFF
--- a/docs/doxygen/include/size_table.md
+++ b/docs/doxygen/include/size_table.md
@@ -9,8 +9,8 @@
     </tr>
     <tr>
         <td>sigv4.c</td>
-        <td><center>4.9K</center></td>
-        <td><center>4.2K</center></td>
+        <td><center>5.1K</center></td>
+        <td><center>4.3K</center></td>
     </tr>
     <tr>
         <td>sigv4_quicksort.c</td>
@@ -19,7 +19,7 @@
     </tr>
     <tr>
         <td><b>Total estimates</b></td>
-        <td><b><center>5.3K</center></b></td>
-        <td><b><center>4.5K</center></b></td>
+        <td><b><center>5.5K</center></b></td>
+        <td><b><center>4.6K</center></b></td>
     </tr>
 </table>
--- a/source/include/sigv4.h
+++ b/source/include/sigv4.h
@@ -122,11 +122,11 @@
 /**
  * @ingroup sigv4_canonical_flags
  * @brief Set this flag to indicate that the HTTP request path, query, and
- * headers are all already canonicalized and request payload is already hashed.
+ * headers are all already canonicalized.
  *
  * This flag is valid only for #SigV4HttpParameters_t.flags.
  */
-#define SIGV4_HTTP_ALL_ARE_CANONICAL_FLAG        0xFU
+#define SIGV4_HTTP_ALL_ARE_CANONICAL_FLAG        0x7U
 
 /**
  * @ingroup sigv4_enum_types
--- a/source/include/sigv4.h
+++ b/source/include/sigv4.h
@@ -122,7 +122,7 @@
 /**
  * @ingroup sigv4_canonical_flags
  * @brief Set this flag to indicate that the HTTP request path, query, and
- * headers are all already canonicalized. and request payload is already hashed.
+ * headers are all already canonicalized and request payload is already hashed.
  *
  * This flag is valid only for #SigV4HttpParameters_t.flags.
  */
--- a/source/include/sigv4.h
+++ b/source/include/sigv4.h
@@ -112,12 +112,21 @@
 
 /**
  * @ingroup sigv4_canonical_flags
- * @brief Set this flag to indicate that the HTTP request path, query, and
- * headers are all already canonicalized.
+ * @brief Set this flag to indicate that the HTTP request payload is
+ * already hashed.
  *
  * This flag is valid only for #SigV4HttpParameters_t.flags.
  */
-#define SIGV4_HTTP_ALL_ARE_CANONICAL_FLAG        0x7U
+#define SIGV4_HTTP_PAYLOAD_IS_HASH               0x8U
+
+/**
+ * @ingroup sigv4_canonical_flags
+ * @brief Set this flag to indicate that the HTTP request path, query, and
+ * headers are all already canonicalized. and request payload is already hashed.
+ *
+ * This flag is valid only for #SigV4HttpParameters_t.flags.
+ */
+#define SIGV4_HTTP_ALL_ARE_CANONICAL_FLAG        0xFU
 
 /**
  * @ingroup sigv4_enum_types
--- a/source/include/sigv4.h
+++ b/source/include/sigv4.h
@@ -55,21 +55,22 @@
 /** @addtogroup sigv4_constants
  *  @{
  */
-#define SIGV4_AWS4_HMAC_SHA256                      "AWS4-HMAC-SHA256"                        /**< AWS identifier for SHA256 signing algorithm. */
-#define SIGV4_AWS4_HMAC_SHA256_LENGTH               ( sizeof( SIGV4_AWS4_HMAC_SHA256 ) - 1U ) /**< AWS identifier for SHA256 signing algorithm. */
-#define SIGV4_HTTP_X_AMZ_DATE_HEADER                "x-amz-date"                              /**< AWS identifier for HTTP date header. */
-#define SIGV4_HTTP_X_AMZ_SECURITY_TOKEN_HEADER      "x-amz-security-token"                    /**< AWS identifier for security token. */
+#define SIGV4_AWS4_HMAC_SHA256                        "AWS4-HMAC-SHA256"                        /**< AWS identifier for SHA256 signing algorithm. */
+#define SIGV4_AWS4_HMAC_SHA256_LENGTH                 ( sizeof( SIGV4_AWS4_HMAC_SHA256 ) - 1U ) /**< AWS identifier for SHA256 signing algorithm. */
+#define SIGV4_HTTP_X_AMZ_DATE_HEADER                  "x-amz-date"                              /**< AWS identifier for HTTP date header. */
+#define SIGV4_HTTP_X_AMZ_SECURITY_TOKEN_HEADER        "x-amz-security-token"                    /**< AWS identifier for security token. */
 
-#define SIGV4_STREAMING_AWS4_HMAC_SHA256_PAYLOAD    "STREAMING-AWS4-HMAC-SHA256-PAYLOAD"      /**< S3 identifier for chunked payloads. */
-#define SIGV4_HTTP_X_AMZ_CONTENT_SHA256_HEADER      "x-amz-content-sha256"                    /**< S3 identifier for streaming requests. */
-#define SIGV4_HTTP_X_AMZ_STORAGE_CLASS_HEADER       "x-amz-storage-class"                     /**< S3 identifier for reduced streaming redundancy. */
+#define SIGV4_STREAMING_AWS4_HMAC_SHA256_PAYLOAD      "STREAMING-AWS4-HMAC-SHA256-PAYLOAD"      /**< S3 identifier for chunked payloads. */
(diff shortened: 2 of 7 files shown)


source/include/sigv4.h
```diff
@@ -110,14 +110,23 @@
...
 /**
  * @ingroup sigv4_canonical_flags
  * @brief Set this flag to indicate that the HTTP request path, query, and
- * headers are all already canonicalized.
+ * headers are all already canonicalized and request payload is already hashed.
  *
  * This flag is valid only for #SigV4HttpParameters_t.flags.
  */
-#define SIGV4_HTTP_ALL_ARE_CANONICAL_FLAG        0x7U
+#define SIGV4_HTTP_ALL_ARE_CANONICAL_FLAG        0xFU
```
muneebahmed10, line 129: Does canonicity have anything to do with being a hash? This looks like it should be staying `0x7U` and the hash flag staying separate

source/sigv4.c
```diff
@@ -2921,6 +2964,38 @@ static SigV4Status_t generateSigningKey( const SigV4Parameters_t * pSigV4Params,
...
+    SigV4Status_t returnStatus = SigV4Success;
+
+    assert( pParams != NULL );
+    assert( pCanonicalContext != NULL );
+
+    if( FLAG_IS_SET( pParams->pHttpParameters->flags, SIGV4_HTTP_PAYLOAD_IS_HASH ) )
+    {
+        /* Copy the hashed payload data supplied by the user in the headers data list. */
+        returnStatus = copyHeaderStringToCanonicalBuffer( pCanonicalContext->pHashPayloadLoc, pCanonicalContext->hashPayloadLen, pParams->pHttpParameters->flags, '\n', pCanonicalContext );
+        pCanonicalContext->pBufCur--;
```
muneebahmed10, line 2980: What's the decrement for?

docs/doxygen/include/size_table.md
```diff
@@ -9,8 +9,8 @@
     </tr>
     <tr>
         <td>sigv4.c</td>
-        <td><center>4.9K</center></td>
-        <td><center>4.2K</center></td>
+        <td><center>5.1K</center></td>
+        <td><center>4.3K</center></td>
```
muneebahmed10, line 13: Is the user likely to have already hashed the payload and encoded it in lowercase hex? Is the extra memory usage worth the speedup gained from not having to hash the payload?

source/sigv4.c
gshvang: We don't need new line but the above function adds new line also.

docs/doxygen/include/size_table.md
gshvang: Yeah, In almost all cases the payload will be hash and hex encoded by the library. So I think it is worth.

muneebahmed10: But is that header something the application is expected to have been added already, or would a user expect the library to add the header for them? If it's the latter case, then the user can save themself time by just telling the library to add it for them

gshvang: The application will add that header before calling the Sigv4 API.

muneebahmed10: So it's not almost all cases then, it's every case

source/sigv4.c
```diff
@@ -1583,6 +1607,17 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
     }
 
 /*-----------------------------------------------------------*/
+    static void storeHashedPayloadLocation( size_t headerIndex,
+                                            CanonicalContext_t * canonicalRequest )
+    {
+        assert( canonicalRequest != NULL );
+
+        if( memcmp( canonicalRequest->pHeadersLoc[ headerIndex ].key.pData, ( const char * ) SIGV4_HTTP_X_AMZ_CONTENT_SHA256_HEADER, canonicalRequest->pHeadersLoc[ headerIndex ].key.dataLen ) == 0 )
```
muneebahmed10, line 1615: I think these MISRA violations can be avoided by passing the string literal as a `const char *` and then assigning those to `const uint8_t *` variables
```suggestion
                                            CanonicalContext_t * canonicalRequest,
                                            const char * pAmzSHA256Header )
    {
        assert( canonicalRequest != NULL );
        const uint8_t * pHeaderData = ( const uint8_t * ) canonicalRequest->pHeadersLoc[ headerIndex ].key.pData;
        const uint8_t * pHeaderLiteral = ( const uint8_t * ) pAmzSHA256Header;

        if( memcmp( pHeaderData, pHeaderLiteral, canonicalRequest->pHeadersLoc[ headerIndex ].key.dataLen ) == 0 )
```

test/unit-test/sigv4_config.h
```diff
@@ -76,7 +76,7 @@
  * <b>Default value:</b> `1024`
  */
 #ifndef SIGV4_PROCESSING_BUFFER_LENGTH
-    #define SIGV4_PROCESSING_BUFFER_LENGTH    350
+    #define SIGV4_PROCESSING_BUFFER_LENGTH    428
```
muneebahmed10, line 79: This seems very arbitrary. Why 428?

test/unit-test/sigv4_utest.c
```diff
@@ -77,6 +77,7 @@
 #define HEADERS                                               "Host: iam.amazonaws.com\r\nContent-Type: application/x-www-form-urlencoded; charset=utf-8\r\nX-Amz-Date: "DATE "\r\n\r\n"
 #define HEADERS_LENGTH                                        ( sizeof( HEADERS ) - 1U )
 #define PRECANON_HEADER                                       "content-type:application/json;\nhost:iam.amazonaws.com\n"
+#define HEADERS_WITH_X_AMZ_CONTENT_SHA256                     "Host: iam.amazonaws.com\r\nContent-Type: application/x-www-form-urlencoded; charset=utf-8\r\nx-amz-content-sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b785\r\nX-Amz-Date: "DATE "\r\n\r\n"
```
muneebahmed10, line 80: This isn't a valid hash. I'd expect the test to fail in this case instead of returning success
```suggestion
#define HEADERS_WITH_X_AMZ_CONTENT_SHA256                     "Host: iam.amazonaws.com\r\nContent-Type: application/x-www-form-urlencoded; charset=utf-8\r\nx-amz-content-sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855\r\nX-Amz-Date: "DATE "\r\n\r\n"
```

source/sigv4.c
```diff
@@ -1681,6 +1726,10 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
                 dataLen = pCurrLoc - pKeyOrValStartLoc;
                 canonicalRequest->pHeadersLoc[ noOfHeaders ].value.pData = pKeyOrValStartLoc;
                 canonicalRequest->pHeadersLoc[ noOfHeaders ].value.dataLen = ( size_t ) dataLen;
+
+                /* Storing location of hashed request payload */
+                storeHashedPayloadLocation( noOfHeaders, ( const char * ) SIGV4_HTTP_X_AMZ_CONTENT_SHA256_HEADER, canonicalRequest );
```
muneebahmed10, line 1731: Why does this need a cast?
```suggestion
                storeHashedPayloadLocation( noOfHeaders, SIGV4_HTTP_X_AMZ_CONTENT_SHA256_HEADER, canonicalRequest );
```

test/unit-test/sigv4_config.h
gshvang: I changed the buffer for the new header `X_AMZ_CONTENT_SHA256` included in the canonical and signed headers.

source/include/sigv4_internal.h
```diff
@@ -205,6 +205,8 @@ typedef struct CanonicalContext
     uint8_t pBufProcessing[ SIGV4_PROCESSING_BUFFER_LENGTH ];       /**< Internal calculation buffer used during canonicalization. */
     char * pBufCur;                                                 /**< pBufProcessing cursor. */
     size_t bufRemaining;                                            /**< pBufProcessing value used during internal calculation. */
+    const char * pHashPayloadLoc;                                   /**< Pointer used to store the location of HTTP request hashed payload. */
```
yourslab, line 208:
```suggestion
    const char * pHashPayloadLoc;                                   /**< Pointer used to store the location of hashed HTTP request payload. */
```

source/include/sigv4_internal.h
```diff
@@ -205,6 +205,8 @@ typedef struct CanonicalContext
...
+    size_t hashPayloadLen;                                          /**< Length of HTTP request hashed payload. */
```
yourslab, line 209:
```suggestion
    size_t hashPayloadLen;                                          /**< Length of hashed HTTP request payload. */
```

source/sigv4.c
```diff
@@ -192,6 +192,18 @@ static SigV4Status_t appendCanonicalizedHeaders( size_t headerCount,
...
+ * @brief Store the location of HTTP request hashed payload in the HTTP request.
+ *
+ * @param[in] headerIndex Index of request Header in the list of parsed headers.
+ * @param[in] pAmzSHA256Header Literal for x-amz-content-sha256 header in HTTP request.
+ * @param[in,out] canonicalRequest Struct to maintain intermediary buffer
+ * and state of canonicalization.
+ */
+static void storeHashedPayloadLocation( size_t headerIndex,
+                                        const char * pAmzSHA256Header,
+                                        CanonicalContext_t * canonicalRequest );
```
yourslab, line 205:
```suggestion
                                        CanonicalContext_t * pCanonicalRequest );
```

test/unit-test/sigv4_utest.c
gshvang: We are checking the coverage here and not checking whether the signature generated matches the one generated at AWS end too.

source/sigv4.c
```diff
@@ -1583,6 +1609,21 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
...
 
 /*-----------------------------------------------------------*/
+    static void storeHashedPayloadLocation( size_t headerIndex,
+                                            const char * pAmzSHA256Header,
+                                            CanonicalContext_t * canonicalRequest )
+    {
+        assert( canonicalRequest != NULL );
+
+        const uint8_t * pHeaderData = ( const uint8_t * ) canonicalRequest->pHeadersLoc[ headerIndex ].key.pData;
+        const uint8_t * pHeaderLiteral = ( const uint8_t * ) pAmzSHA256Header;
```
yourslab, line 1619: Do we need an intermediate `pHeaderLiteral` variable here? We can just use `pAmzSHA256Header` and cast directly?

yourslab: Or better yet, just use the macro directly

test/unit-test/sigv4_utest.c
muneebahmed10: We should still avoid using invalid parameters in tests when possible. If the implementation is changed to verify the length of the hash, then this test will start failing for a reason unrelated to what it's meant to test for

source/sigv4.c
```diff
@@ -1583,6 +1609,21 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
...
+
+        if( memcmp( pHeaderData, pHeaderLiteral, canonicalRequest->pHeadersLoc[ headerIndex ].key.dataLen ) == 0 )
```
yourslab, line 1621: What happens if `dataLen` is a prefix of `pAmzSHA256Header`? I think this still evaluate to true, which we don't want to happen.

source/sigv4.c
muneebahmed10: Yes we do. MISRA reasons

gshvang: This is done to avoid MISRA violation(doing the cast within the function call still triggers 21.16.)

muneebahmed10 (approved):

yourslab (approved):

yourslab:
The CBMC proofs are still failing. I suggest running them locally.

gshvang:
> The CBMC proofs are still failing. I suggest running them locally.

Yeah, I am investigating that and will raise a separate PR for it.

dan4thewin (approved):

PR #52, opened 2021-08-18
Bug fixes in library and hygiene improvements in test cases

aggarw13:
### Hygiene Changes
* Address unresolved feedback from PR #29 
* Improve tests to include verification of generated signature in happy-path test cases 

### Bugfixes
Fix 2 categories of bugs found from adding signature verification in tests: 
1. Fix parsing of query string for cases **empty parameter values** (with the format `<param>=&<param2>` OR `<param>&<param2>`) 
2. Fix URI encoding to not encode `NULL` character.

### DIFF
--- a/source/sigv4.c
+++ b/source/sigv4.c
@@ -2494,8 +2494,7 @@ static int32_t hmacFinal( HmacContext_t * pHmacContext,
          * this is by performing XOR on each byte of the inner-padded key (ipad ^ opad).  */
         for( i = 0U; i < pCryptoInterface->hashBlockLen; i++ )
         {
-            static const uint8_t padByteForOuterKey = HMAC_INNER_PAD_BYTE ^ HMAC_OUTER_PAD_BYTE;
-            pHmacContext->key[ i ] ^= padByteForOuterKey;
+            pHmacContext->key[ i ] ^= HMAX_IPAD_XOR_OPAD_BYTE;
         }
 
         returnStatus = pCryptoInterface->hashInit( pCryptoInterface->pHashContext );
--- a/source/sigv4.c
+++ b/source/sigv4.c
@@ -1819,13 +1819,73 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
 
 /*-----------------------------------------------------------*/
 
-    static SigV4Status_t setQueryStringFieldsAndValues( const char * pQuery,
-                                                        size_t queryLen,
-                                                        size_t * pNumberOfParameters,
-                                                        CanonicalContext_t * pCanonicalRequest )
+    static void processAmpersantInQueryString( bool fieldHasValue,
+                                               size_t currQueryIndex,
+                                               size_t startOfFieldOrValue,
+                                               size_t * pCurrParamCount,
+                                               const char * pQuery,
+                                               CanonicalContext_t * pCanonicalRequest )
+    {
+        bool previousParamIsValid = true;
+
+        assert( pCurrParamCount != NULL );
+        assert( pQuery != NULL );
+        assert( pCanonicalRequest != NULL );
+
+        /* Ignore unnecessary '&' that represent empty parameter names.
+         * Trimmable '&'s can be leading, trailing or repeated as separators between paramater
+         * pairs.
+         * For example, this function will prune "&p1=v1&&p2=v2&" to "p1=v1&p2=v2". */
+        if( !( fieldHasValue ) && ( ( ( currQueryIndex - startOfFieldOrValue ) == 0U ) ) )
+        {
+            /* Do nothing as previous parameter name is empty. */
+            previousParamIsValid = false;
+        }
+
+        /* Case when parameter has empty value even though the query parameter entry has the form
+         * "<param>=". */
+        else if( ( fieldHasValue ) && ( ( currQueryIndex - startOfFieldOrValue ) == 0U ) )
+        {
+            /* Store the previous parameter's empty value information. Use NULL to represent empty value. */
+            setQueryParameterValue( *pCurrParamCount, NULL, 0U, pCanonicalRequest );
+        }
+
+        /* Case when a new query parameter pair has begun without the previous parameter having an associated value,
+         * i.e. of the format "<param1>&<param2>=<value>" where "<param1>".
+         * In thus case, as the previous parameter does not have a value, we will store an empty value for it. */
+        else if( !( fieldHasValue ) )
+        {
+            /* Store information about previous query parameter name. The query parameter has no associated value. */
+            setQueryParameterKey( *pCurrParamCount, &pQuery[ startOfFieldOrValue ], currQueryIndex - startOfFieldOrValue, pCanonicalRequest );
+
+            /* Store the previous parameter's empty value information. Use NULL to represent empty value. */
+            setQueryParameterValue( *pCurrParamCount, NULL, 0U, pCanonicalRequest );
+        }
+        /* This represents the case when the previous parameter has a value associated with it. */
+        else
+        {
+            /* End of value reached, so store a pointer to the previously set value. */
+            setQueryParameterValue( *pCurrParamCount, &pQuery[ startOfFieldOrValue ], currQueryIndex - startOfFieldOrValue, pCanonicalRequest );
+        }
+
+        if( previousParamIsValid == true )
+        {
+            /* As the previous parameter is valid, increment the total parameters
+             * parsed so far from the query string. */
+            ( *pCurrParamCount )++;
+        }
+    }
+
+/*-----------------------------------------------------------*/
+
+    static SigV4Status_t parseQueryString( const char * pQuery,
+                                           size_t queryLen,
+                                           size_t * pNumberOfParameters,
+                                           CanonicalContext_t * pCanonicalRequest )
     {
         size_t currentParameter = 0U, i = 0U, startOfFieldOrValue = 0U;
         bool fieldHasValue = false;
+        bool fieldFound = false;
         SigV4Status_t returnStatus = SigV4Success;
 
         assert( pNumberOfParameters != NULL );
@@ -1835,7 +1895,10 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
         /* Note: Constness of the query string is casted out here, taking care not to modify
          * its contents in any way. */
 
-        /* Set cursors to each field and value in the query string. */
+        /* Set cursors to each field and value in the query string.
+         * Note: We iterate the index to queryLen (instead of queryLen - 1) for the
+         * special case of handling the last index as a terminating value of a query
+         * parameter. */
         for( i = 0U; i <= queryLen; i++ )
         {
             /* This test is at the beginning of the loop to ensure that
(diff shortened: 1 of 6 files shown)


source/sigv4.c
```diff
@@ -1814,10 +1819,69 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
...
+
+        assert( pCurrParamCount != NULL );
+        assert( pQuery != NULL );
+        assert( pCanonicalRequest != NULL );
+
+        /* Ignore unnecessary '&' that represent empty parameter names.
+         * Trimmable '&'s can be leading, trailing or repeated as separators between parameter
+         * pairs.
+         * For example, this function will prune "&p1=v1&&p2=v2&" to "p1=v1&p2=v2". */
+        if( !( fieldHasValue ) && ( ( ( currQueryIndex - startOfFieldOrValue ) == 0U ) ) )
```
muneebahmed10, line 1839: If there's a trimmable `&`, would `startOfFieldOrValue` need to be incremented?

source/sigv4.c
```diff
@@ -1814,10 +1819,69 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
 
 /*-----------------------------------------------------------*/
 
-    static SigV4Status_t setQueryStringFieldsAndValues( const char * pQuery,
-                                                        size_t queryLen,
-                                                        size_t * pNumberOfParameters,
-                                                        CanonicalContext_t * pCanonicalRequest )
+    static void processAmpersantInQueryString( bool fieldHasValue,
```
muneebahmed10, line 1822:
```suggestion
    static void processAmpersandInQueryString( bool fieldHasValue,
```

source/sigv4.c
```diff
@@ -1814,10 +1819,69 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
...
+         * For example, this function will prune "&p1=v1&&p2=v2&" to "p1=v1&p2=v2". */
+        if( !( fieldHasValue ) && ( ( ( currQueryIndex - startOfFieldOrValue ) == 0U ) ) )
+        {
+            /* Do nothing as previous parameter name is empty. */
+            previousParamIsValid = false;
+        }
+
+        /* Case when parameter has empty value even though the query parameter entry has the form
+         * "<param>=". */
+        else if( ( fieldHasValue ) && ( ( currQueryIndex - startOfFieldOrValue ) == 0U ) )
```
muneebahmed10, line 1847: So this will happen in the case of either `<param>=<end of query>` or `<param>=&`. Are either of these valid?

source/sigv4.c
```diff
@@ -1814,10 +1819,69 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
...
+        {
+            /* Store the previous parameter's empty value information. Use NULL to represent empty value. */
+            setQueryParameterValue( *pCurrParamCount, NULL, 0U, pCanonicalRequest );
+        }
+
+        /* Case when a new query parameter pair has begun without the previous parameter having an associated value,
+         * i.e. of the format "<param1>&<param2>=<value>" where "<param1>".
+         * In thus case, as the previous parameter does not have a value, we will store an empty value for it. */
+        else if( !( fieldHasValue ) )
```
muneebahmed10, line 1856: You can combine either the first `if` and `else if` into a single `if( ( currQueryIndex - startOfFieldOrValue ) == 0U )`, or the first `if` and this `else if ` into a single `if( !fieldHasValue )`

source/sigv4.c
```diff
@@ -1972,53 +2021,50 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
...
+                                      true /* Encode slash (/) */,
+                                      false /* Do not encode '='. */ );
 
             if( returnStatus == SigV4Success )
             {
                 pBufLoc += encodedLen;
                 remainingLen -= encodedLen;
 
-                /* Encode parameter value if non-empty. Query parameters can have empty values. */
-                if( pCanonicalRequest->pQueryLoc[ paramsIndex ].value.dataLen > 0U )
```
muneebahmed10, line 1984: Why can this be assumed to be true now? Or does `writeValueInCanonicalizedQueryString` still work with a zero length parameter?

source/sigv4.c
```diff
@@ -2434,7 +2481,8 @@ static int32_t hmacFinal( HmacContext_t * pHmacContext,
          * this is by performing XOR on each byte of the inner-padded key (ipad ^ opad).  */
         for( i = 0U; i < pCryptoInterface->hashBlockLen; i++ )
         {
-            pHmacContext->key[ i ] ^= ( HMAC_INNER_PAD_BYTE ^ HMAC_OUTER_PAD_BYTE );
+            static const uint8_t padByteForOuterKey = HMAC_INNER_PAD_BYTE ^ HMAC_OUTER_PAD_BYTE;
```
muneebahmed10, line 2484: Is adding a new const and relying on the compiler to optimize it for you better than just defining a macro that's 0x6A?

test/unit-test/sigv4_utest.c
```diff
@@ -62,7 +62,8 @@
 #define QUERY_WITH_SPECIAL_CHARS                              "param=/"
 
 #define QUERY_STRING_NO_PARAM_VALUE                           "param=&param2="
-#define QUERY_STRING_STARTING_WITH_AMPERSANT                  "&param2="
+#define QUERY_STRING_WITH_TRAILING_N_LEADING_AMPERSANTS       "&&param2=&&"
+#define QUERY_STRING_WITH_REPEATED_AMPERSANTS                 "param1=val&&param2=val"
```
muneebahmed10, line 66:
```suggestion
#define QUERY_STRING_WITH_TRAILING_N_LEADING_AMPERSANDS       "&&param2=&&"
#define QUERY_STRING_WITH_REPEATED_AMPERSANDS                 "param1=val&&param2=val"
```

test/unit-test/sigv4_utest.c
```diff
@@ -625,55 +626,83 @@ void test_SigV4_GenerateHTTPAuthorization_Invalid_Params()
...
+    returnStatus = SigV4_GenerateHTTPAuthorization( &params, authBuf, &authBufLen, &signature, &signatureLen );
+    TEST_ASSERT_EQUAL( SigV4InvalidParameter, returnStatus );
 }
 
 /* TODO - Verify the generated signatures. */
 void test_SigV4_GenerateHTTPAuthorization_Happy_Paths()
 {
     SigV4Status_t returnStatus;
 
+    const char * pExpectedSignature = "20fdb62349e7104f9ce4184a444fedfbd19e40a5e31d57d433689c5a5138fa99";
```
muneebahmed10, line 642: So can the TODO on line 637 be removed?

source/sigv4.c
aggarw13: The `startOfFieldOrValue` value is always incremented in the calling code of this function. I will move that logic to this function to easily notice that.

muneebahmed10: Didn't notice that that was changed. It looks fine as is then, as it looks like `startOfFieldOrValue` should always be incremented when encountering an `&`, especially for the case of "&&" which wasn't handled previously.

source/sigv4.c
aggarw13: The [API Gateway Tutorial](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-step-by-step.html#getting-started-mappings) provides an example that supports providing query parameters with no value in both the formats. 
The example REST API is: http://petstore-demo-endpoint.execute-api.com/petstore/pets?type=cat&page=2

aggarw13: Moreover, the [AWS SigV4 spec](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html) mentions about providing the query parameter name with `=` character but an empty string for empty values.

> For each parameter, append the URI-encoded parameter name, followed by the equals sign character (=), followed by the URI-encoded parameter value. Use an empty string for parameters that have no value.

source/sigv4.c
aggarw13: >Or does writeValueInCanonicalizedQueryString still work with a zero length parameter?

Yes, this function adds the `=` character which is required for canonical query format even with empty values.

source/sigv4.c
aggarw13: I'm not sure that I follow your suggestion. All the four cases are distinct with separate handling. I have changed the if-else chain to instead use a nested if-else to be based off of the `fieldHasValue` data to make it easier to follow.

source/sigv4.c
aggarw13: I stylistically prefer storing only publicly known information about the algorithm/protocol in the collection of macro constants in the header file. 
Wouldn't use of `static const` guarantee that the value is computed only at static initialization?

test/unit-test/sigv4_utest.c
```diff
@@ -62,7 +62,8 @@
 #define QUERY_WITH_SPECIAL_CHARS                              "param=/"
 
 #define QUERY_STRING_NO_PARAM_VALUE                           "param=&param2="
-#define QUERY_STRING_STARTING_WITH_AMPERSANT                  "&param2="
+#define QUERY_STRING_WITH_TRAILING_N_LEADING_AMPERSANT        "&&param2=&&"
+#define QUERY_STRING_WITH_REPEATED_AMPERSANT                  "param1=val&&param2=val"
```
muneebahmed10, line 66: Here too
```suggestion
#define QUERY_STRING_WITH_TRAILING_N_LEADING_AMPERSAND        "&&param2=&&"
#define QUERY_STRING_WITH_REPEATED_AMPERSAND                  "param1=val&&param2=val"
```

source/sigv4.c
muneebahmed10: Just thought it looked odd seeing a constant here instead of a macro. I'm not sure what's so different about 0x6A. It's just the XOR of 0x36 and 0x5C, the two constants defined in the HMAC RFC.

muneebahmed10 (dismissed):
Some spelling

source/sigv4.c
muneebahmed10: Yep, that's what I meant

muneebahmed10 (approved):

source/sigv4.c
aggarw13: Okay, I have added a macro constant for consistency with use of macros for constants

dan4thewin (approved):

source/sigv4.c
```diff
@@ -1972,53 +2034,50 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
...
-                    assert( pCanonicalRequest->pQueryLoc[ paramsIndex ].value.pData != NULL );
-                    returnStatus = writeValueInCanonicalizedQueryString( pBufLoc,
-                                                                         remainingLen,
-                                                                         pCanonicalRequest->pQueryLoc[ paramsIndex ].value.pData,
-                                                                         pCanonicalRequest->pQueryLoc[ paramsIndex ].value.dataLen,
-                                                                         &encodedLen );
-                    pBufLoc += encodedLen;
-                    remainingLen -= encodedLen;
-                }
+                assert( pCanonicalRequest->pQueryLoc[ paramsIndex ].value.pData != NULL );
```
digantaTheProgrammer, line 2045: @aggarw13 Since, as you mentioned above that, `writeValueInCanonicalizedQueryString` can work with empty values, why are we asserting this?

Doesn't it essentially force all parameters to have non-empty value?

PR #59, opened 2021-08-31
Remove superfluous memmove

yourslab:
The [existing SigV4 library has a copy operation](https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/blob/v1.0.0/source/sigv4.c#L2604-L2610) in the internal processing buffer while generating the **String to Sign**. Instead of copying the hex encoded bytes of the Canonical Request data from a mid-location (right after the initial presence of the Canonical Request) to a location where the _string-to-sign_ data can be stored in the beginning of the buffer, the copy operation can be avoided by storing the _string-to-sign_ at the end of the processing buffer which will cause the processing buffer to contain information in of the **String-To-Sign** and **Signing Key** in the `{<signing-key>, <string-to-sign>}` order

### DIFF
--- a/source/sigv4.c
+++ b/source/sigv4.c
@@ -2718,6 +2718,7 @@ static SigV4Status_t writeStringToSign( const SigV4Parameters_t * pParams,
     ptrdiff_t bufferLen = pCanonicalContext->pBufCur - pBufStart;
     /* An overestimate but sufficient memory is checked before proceeding. */
     size_t encodedLen = SIGV4_PROCESSING_BUFFER_LENGTH;
+
     /* The string to sign is composed of (+ means string concatenation):
      * Algorithm + \n +
      * RequestDateTime + \n +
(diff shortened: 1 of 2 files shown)


source/sigv4.c
```diff
@@ -2713,50 +2714,45 @@ static SigV4Status_t writeStringToSign( const SigV4Parameters_t * pParams,
                                         CanonicalContext_t * pCanonicalContext )
 {
     SigV4Status_t returnStatus = SigV4Success;
-    size_t encodedLen = pCanonicalContext->bufRemaining;
     char * pBufStart = ( char * ) pCanonicalContext->pBufProcessing;
     ptrdiff_t bufferLen = pCanonicalContext->pBufCur - pBufStart;
+    /* An overestimate but sufficient memory is checked before proceeding. */
+    size_t encodedLen = SIGV4_PROCESSING_BUFFER_LENGTH;
+    size_t sizeNeededBeforeHash = algorithmLen + 1U + \
```
muneebahmed10, line 2721: Could be helpful to comment what each part before the hash is. Are the 1s for newline characters?

muneebahmed10 (approved):

muneebahmed10 (approved):

paulbartell (approved):

PR #71, opened 2022-08-24
MISRA Compliance Updates

Skptak:
Updated the MISRA.md file to match the new format for it.
Added the 8.7 rule to the misra.config file.
Small changes to casts in an if statement and an addition/subtraction
for MISRA compliance.
Added in the DNDEBUG and DISABLE_LOGGING when building the
coverity_analysis target.
Added coverity to the lexicon

### DIFF
--- a/MISRA.md
+++ b/MISRA.md
@@ -1,23 +1,22 @@
 # MISRA Compliance
 
-The AWS SigV4 Library files conform to the
-[MISRA C:2012](https://www.misra.org.uk)
+The AWS SigV4 Library files conform to the [MISRA C:2012](https://www.misra.org.uk)
 guidelines, with some noted exceptions. Compliance is checked with Coverity static analysis.
-Deviations from the MISRA standard are listed below:
+The specific deviations, suppressed inline, are listed below.
 
-### Ignored by [Coverity Configuration](tools/coverity/misra.config)
-| Deviation | Category | Justification |
-| :-: | :-: | :-: |
-| Directive 4.9 | Advisory | Allow inclusion of function like macros. Asserts and logging macros use function like macros. |
-| Rule 2.4 | Advisory | Allow unused tags. Some compilers warn if types are not tagged. |
-| Rule 2.5 | Advisory | Allow unused macros. |
-| Rule 3.1 | Required | Allow nested comments. C++ style `//` comments are used in example code within Doxygen documentation blocks. |
-| Rule 11.5 | Advisory | Allow casts from `void *`. Functions with `void *` parameters are used while sorting. |
-
-### Flagged by Coverity
-| Deviation | Category | Justification |
-| :-: | :-: | :-: |
-| Rule 8.7 | Advisory | API functions are not used by the library outside of the files they are defined; however, they must be externally visible in order to be used by an application. |
+Additionally, [MISRA configuration file](https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/blob/main/tools/coverity/misra.config) contains the project wide deviations.
 
 ### Suppressed with Coverity Comments
-*None.*
+To find the violation references in the source files run grep on the source code
+with ( Assuming rule 11.4 violation; with justification in point 2 ):
+```
+grep 'MISRA Ref 11.4.2' . -rI
+```
+
+#### Rule 5.4
+_Ref 5.4-1_
+
+- MISRA Rule 5.4 flags the following macro's name as ambiguous from the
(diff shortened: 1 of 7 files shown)


source/sigv4.c
```diff
@@ -1508,9 +1508,10 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
     {
         char outputChar;
 
+
         if( ( inputChar >= 'A' ) && ( inputChar <= 'Z' ) )
         {
-            outputChar = 'a' + inputChar - 'A';
+            outputChar = ( ( int8_t ) 'a' ) + inputChar - ( ( int8_t ) 'A' );
```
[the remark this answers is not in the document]
Skptak: Came up with a different way of doing this, have updated the line.
For potential future reference however: MISRA/Coverity can run into issues when dealing with math involving chars. From what I can tell the issue comes from elevating from a char to a larger int when doing the math. Casts to int8_t get it around this.

jasonpcarroll (approved):

MISRA.md
```diff
@@ -1,23 +1,22 @@
...
-
-### Flagged by Coverity
-| Deviation | Category | Justification |
-| :-: | :-: | :-: |
-| Rule 8.7 | Advisory | API functions are not used by the library outside of the files they are defined; however, they must be externally visible in order to be used by an application. |
+Additionally, [MISRA configuration file](https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/blob/main/tools/coverity/misra.config) contains the project wide deviations.
 
 ### Suppressed with Coverity Comments
-*None.*
+To find the violation references in the source files run grep on the source code
```
AniruddhaKanhere, line 10:
```suggestion
To find the deviation references in the source files run grep on the source code
```

MISRA.md
```diff
@@ -1,23 +1,22 @@
...
+```
+grep 'MISRA Ref 11.4.2' . -rI
+```
+
+#### Rule 5.4
+_Ref 5.4-1_
+
+- MISRA Rule 5.4 flags the following macro's name as ambiguous from the
+        one postfixed with _LEN. While it would be possible to change this name to
+        satisify this rule. We have chosen to suppress this singular occurence for
```
AniruddhaKanhere, line 21:
```suggestion
        satisfy this rule. We have chosen to suppress this singular occurrence for
```

MISRA.md
```diff
@@ -1,23 +1,22 @@
...
+with ( Assuming rule 11.4 violation; with justification in point 2 ):
+```
+grep 'MISRA Ref 11.4.2' . -rI
+```
+
+#### Rule 5.4
+_Ref 5.4-1_
+
+- MISRA Rule 5.4 flags the following macro's name as ambiguous from the
+        one postfixed with _LEN. While it would be possible to change this name to
```
AniruddhaKanhere, line 20: I think this one should be:
```suggestion
        one postfixed with _LENGTH. While it would be possible to change this name to
```

test/include/sigv4_config.h
```diff
@@ -30,12 +30,30 @@
 
 #include <stdio.h>
 
-#define LogError( message )    printf( "Error: " ); printf message; printf( "\n" )
+#ifdef DISABLE_LOGGING
```
AniruddhaKanhere, line 33: Why do we need this if we are using NDEBUG as well?
Does this serve some different purpose?

tools/coverity/misra.config
```diff
@@ -21,9 +21,13 @@
             deviation: "Rule 3.1",
             reason: "Allow nested comments. C++ style `//` comments are used in example code within Doxygen documentation blocks."
         },
+        {
+            deviation: "Rule 8.7",
+            reason: "API functions are not used by the library outside of the files they are defined; however, they must be externally visible in orde    r to be used by an application."
+        },
```
AniruddhaKanhere, line 27:
```suggestion
            reason: "API functions are not used by the library outside of the files they are defined in; however, they must be externally visible in order to be used by an application."
        },
```

AniruddhaKanhere (changes requested):

MISRA.md
Skptak: Put this change in, is this something we want to open a set of PR's for in the other libraries or just leave it as is for now?

AniruddhaKanhere: I think we should have caught this earlier. For consistency, we should stick to either violation or deviation. But, I think it is not a blocker.

jasonpcarroll (approved):

PR #85, opened 2023-06-26
fix array index off-by-one error in canonical parameter value generation

schoeler:
Hi,
I noticed the canonical generation of http query parameter values were not converting the special characters correctly.  This should fix that.  I hope you can add tests around this use case in the future.
Thanks,
Ben Schoeler
Fortune Brands Home & Security


archigup:
Looks good, thanks for contributing!

archigup (approved):

johnrhen (approved):

PR #92, opened 2024-02-27
Fix MISRA C 2012 deviations

chinglee-iot:
In this PR
* Suppress rule 18.2 deviations with comment. The pointer addresses elements of the same array.
* Fix rule 18.4 deviations. Not to apply operators to an expression of pointer type.
* Fix rule 21.2 deviations. Not to use reserved identifier in internal variables.


moninom1 (approved):

source/sigv4.c
```diff
@@ -1115,27 +1115,27 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
...
-    pBufWrite += pSigV4Params->serviceLen;
+    pBufWrite = &pBufWrite[ pSigV4Params->serviceLen ];
 
     *pBufWrite = CREDENTIAL_SCOPE_SEPARATOR;
-    pBufWrite += CREDENTIAL_SCOPE_SEPARATOR_LEN;
+    pBufWrite = &pBufWrite[ CREDENTIAL_SCOPE_SEPARATOR_LEN ];
 
     /* Concatenate terminator. */
-    pBufWrite += copyString( pBufWrite, CREDENTIAL_SCOPE_TERMINATOR, CREDENTIAL_SCOPE_TERMINATOR_LEN );
+    pBufWrite = &pBufWrite[ copyString( pBufWrite, CREDENTIAL_SCOPE_TERMINATOR, CREDENTIAL_SCOPE_TERMINATOR_LEN ) ];
```
AniruddhaKanhere, line 1138: Can we split this into 2 statements?

```suggestion
    size_t result = copyString( pBufWrite, CREDENTIAL_SCOPE_TERMINATOR, CREDENTIAL_SCOPE_TERMINATOR_LEN );
    pBufWrite = &pBufWrite[ result ];
```

source/sigv4.c
```diff
@@ -2653,10 +2655,10 @@ static SigV4Status_t writeLineToCanonicalRequest( const char * pLine,
         ( void ) memcpy( pCanonicalContext->pBufCur,
                          pLine,
                          lineLen );
-        pCanonicalContext->pBufCur += lineLen;
+        pCanonicalContext->pBufCur = &pCanonicalContext->pBufCur[ lineLen ];
```
AniruddhaKanhere, line 2658:
```suggestion
        pCanonicalContext->pBufCur = &( pCanonicalContext->pBufCur[ lineLen ] );
```

source/sigv4.c
```diff
@@ -2653,10 +2655,10 @@ static SigV4Status_t writeLineToCanonicalRequest( const char * pLine,
...
 
         *( pCanonicalContext->pBufCur ) = LINEFEED_CHAR;
-        pCanonicalContext->pBufCur += 1U;
+        pCanonicalContext->pBufCur = &pCanonicalContext->pBufCur[ 1 ];
```
AniruddhaKanhere, line 2661:
```suggestion
        pCanonicalContext->pBufCur = &( pCanonicalContext->pBufCur[ 1 ] );
```

source/sigv4.c
```diff
@@ -3291,6 +3293,8 @@ SigV4Status_t SigV4_GenerateHTTPAuthorization( const SigV4Parameters_t * pParams
      * Note that the StringToSign starts from the beginning of the processing buffer. */
     if( returnStatus == SigV4Success )
     {
+        /* More details at: https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/blob/main/MISRA.md#rule-182 */
```
AniruddhaKanhere, line 3296: Add this line above as well:

```
        /* MISRA Ref 18.2.1 [Pointer subtraction within array] */
```

AniruddhaKanhere: See here for reference: https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/blob/e82835329069d804e0db883d69eb4e1124dd7f08/source/include/sigv4.h#L67

AniruddhaKanhere (commented):
We should also update the README here: https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/blob/main/tools/coverity/README.md to say the actual version being used. Currently is says that:
> Note For generating the report as outlined below, we have used Coverity version 2018.09.

source/sigv4.c
chinglee-iot: Specify operator with parentheses in [this commit](https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/pull/92/commits/14e419b36f427f681805a2c987f0cdf37b2e0263).

source/sigv4.c
chinglee-iot: Specify operator with parentheses in [this commit](https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/pull/92/commits/14e419b36f427f681805a2c987f0cdf37b2e0263).

source/sigv4.c
```diff
@@ -1656,7 +1659,8 @@ static void generateCredentialScope( const SigV4Parameters_t * pSigV4Params,
         if( sigV4Status == SigV4Success )
         {
             /* Replacing the last ';' with '\n' as last header should not have ';'. */
-            *( canonicalRequest->pBufCur - 1 ) = '\n';
+            pBufPtr = ( char * ) ( &canonicalRequest->pBufProcessing[ canonicalRequest->pBufCur - ( char * ) ( &canonicalRequest->pBufProcessing ) - 1 ] );
+            *pBufPtr = '\n';
```
AniruddhaKanhere, line 1663: Can this not be simplified like so?
```suggestion
            canonicalRequest->pBufCur[ -1 ] = '\n';
```

I think it is too hard to understand right now by reading it.

chinglee-iot: Update in new commit .

AniruddhaKanhere (approved):

ericbj29 (approved):

.github/workflows/ci.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           cmake -S test -B build/ \
           -G "Unix Makefiles" \
           -DCMAKE_BUILD_TYPE=Debug \
-          -DBUILD_UNIT_TESTS=ON \
+          -DUNITTEST=ON \
```
archigup, line 37: If we are changing this, might as well change it to BUILD_TESTING which is the standard variable for enabling/disabling testing (see https://cmake.org/cmake/help/latest/module/CTest.html, though its commonly used for testing even if not using ctest).

MISRA.md
```diff
@@ -21,3 +21,13 @@ _Ref 5.4.1_
         in use and changing the name will break existing user projects. Thus, for
         backwards compatibility, the macro is not modified and kept as is and the
         deviation is suppressed.
+
+_Ref 18.2.1_
+
+- MISRA Rule 18.2 states that two pointers may only be subtracted if they point
```
archigup, line 27: Since we have offsets, this should not require pointer arithmetic? Could we convert this to do the math on indices?

test/CMakeLists.txt
```diff
@@ -31,24 +35,26 @@ set( CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib )
 # Include filepaths for source and include.
 include( ${MODULE_ROOT_DIR}/sigv4FilePaths.cmake )
 
-# Target for Coverity analysis that builds the library.
-add_library( coverity_analysis
-            ${SIGV4_SOURCES} )
+if( COV_ANALYSIS )
```
archigup, line 38: Do we need a coverity specific target at all? Coverity should be able to work with the normal targets.

source/sigv4.c
```diff
@@ -3291,6 +3293,9 @@ SigV4Status_t SigV4_GenerateHTTPAuthorization( const SigV4Parameters_t * pParams
      * Note that the StringToSign starts from the beginning of the processing buffer. */
     if( returnStatus == SigV4Success )
     {
+        /* MISRA Ref 18.2.1 [Pointer subtraction within array] */
+        /* More details at: https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/blob/main/MISRA.md#rule-182 */
+        /* coverity[misra_c_2012_rule_18_2_violation] */
         bufferLen = canonicalContext.pBufCur - ( char * ) canonicalContext.pBufProcessing;
```
archigup, line 3299: Could this pBufCur's type be changed from a pointer to an array index?

.github/workflows/ci.yml
chinglee-iot: `BUILD_UNIT_TESTS` is also not a cmake standard variable. This macro is updated to align other FreeRTOS libraries. Since it is only used in FreeRTOS libraries, using `UNITTEST` doesn't seem to have problem.

source/sigv4.c
chinglee-iot: > Rule 18.2 Subtraction between pointers shall only be applied to pointers that
address elements of the same array

For the [header file](https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/blob/dbe5583e50a6fd232f35c728c5d0941e4f012ff1/source/include/sigv4_internal.h#L209), the `pBufCur` is a `pBufProcessing` curosr and addresses `pBufProcessing` elements in the design. This array substraction follows MISRA rule 18.2.

It looks like coverity can't tell if they address elements of the same array. Therefore, I added an exception here. I also added assert here to ensure that pBufCur points to the same array.

test/CMakeLists.txt
chinglee-iot: This is also an update to align FreeRTOS libraries design. It is true that normal targets can work with coverity. However, normal project may not configured properly to compile all the code we want to run coverity with. Another advantage of the coverity project is simplicity and independence.

MISRA.md
chinglee-iot: Please reference [this reply](https://github.com/aws/SigV4-for-AWS-IoT-embedded-sdk/pull/92#discussion_r1510667278).

source/sigv4.c
archigup: Yes, I am suggesting we change that. Using indexes into the buffer instead of pointers into the buffer is probably a better design, and would not trip up coverity.

AniruddhaKanhere: We discussed this and we shall fix the above issue by refactoring in a separate PR. That will keep this PR 'clean'.

AniruddhaKanhere (approved):

archigup (approved):
Approved, given we make the changes to not need the exception afterwards.

PR #107, opened 2025-05-01
Clean up warnings from Coverity 2024.6.1

archigup:


source/sigv4.c
```diff
@@ -3123,6 +3123,7 @@ SigV4Status_t SigV4_GenerateHTTPAuthorization( const SigV4Parameters_t * pParams
                                                size_t * signatureLen )
 {
     SigV4Status_t returnStatus = SigV4Success;
+    /* coverity[stack_use] 10528 bytes on x86_64 */
```
archigup, line 3126: This maybe should not be on the stack, but that can be future work

kstribrnAmzn (approved):

